PR #198, opened 2026-03-23
Sync Redis isolation and process management scripts

08mamba24:
## Summary
- add configurable `REDIS_KEY_PREFIX` support for Redis instance isolation
- fix `ChatContainer` thread/store resync behavior
- add process cleanup scripts and mark related scripts executable

## Notes
- source branch: `08mamba24:main`
- synced from local changes in both `clowder-ai` and `cat-cafe-runtime`


scripts/kill-project-processes.sh
```diff
@@ -0,0 +1,109 @@
...
+}
+
+echo -e "${YELLOW}Scanning for processes related to:${NC} $PROJECT_PATH"
+echo ""
+
+# Find all processes related to the project directory
+# This matches: node, tsx, next, esbuild, pnpm processes that contain the project path
+PIDS=$(ps aux | grep -E "(node|tsx|next|esbuild|pnpm)" | grep -E "$PROJECT_PATH" | grep -v grep | awk '{print $2}' || true)
+
+# Also find related child processes that might be in node_modules but not show full path
```
zts212653: ⚠️ **Linux-only**: `readlink "/proc/$pid/cwd"` only works on Linux. On macOS there is no `/proc` filesystem — you'd need `lsof -p $pid` or similar.

At minimum, add a platform check at the top:
```bash
if [[ "$(uname)" != "Linux" ]]; then
    echo "This script requires Linux (/proc filesystem)"
    exit 1
fi
```

packages/shared/src/utils/redis.ts
```diff
@@ -15,7 +15,7 @@ export interface RedisConfig {
 export function getDefaultRedisConfig(): RedisConfig {
   return {
     url: process.env['REDIS_URL'] ?? 'redis://localhost:6399',
-    keyPrefix: 'cat-cafe:',
+    keyPrefix: process.env['REDIS_KEY_PREFIX'] ?? 'cat-cafe:',
```
zts212653: ✅ This change is clean and backward-compatible. 

One thing to be aware of: `packages/api/src/scripts/migrate-signals/cli.ts` creates Redis clients via `new Redis(url)` directly, bypassing this config function — so the isolation isn't complete yet. See #199 for the full scope.

packages/web/src/components/ChatContainer.tsx
```diff
@@ -269,22 +269,23 @@ export function ChatContainer({ threadId }: ChatContainerProps) {
   const storeThreads = useChatStore((s) => s.threads);
   const prevThreadRef = useRef(threadId);
   useEffect(() => {
-    if (prevThreadRef.current !== threadId) {
-      // Thread switch: store saves/restores per-thread state automatically
+    const storeThreadId = useChatStore.getState().currentThreadId;
```
zts212653: ⚠️ **Needs reproduction steps**: The `useChatStore.getState()` snapshot approach is sound (Zustand's `getState()` is synchronous and won't trigger re-render), but we'd like to understand:

1. What specific user flow causes `storeThreadId !== threadId` when `prevThreadRef.current === threadId`?
2. Can you add a regression test?

This would make the fix much more confident to merge. See #200.

zts212653 (changes requested):
Hi @08mamba24 👋 Thanks for the contribution! These changes clearly come from real multi-instance deployment experience, and the direction is valuable.

We've reviewed this PR from the maintainer perspective (two independent reviewers) and have a few findings. We've also created dedicated issues for each concern to make tracking cleaner:

## Summary

This PR bundles **three independent changes** — Redis key prefix isolation, ChatContainer thread/store resync, and process cleanup scripts. We'd recommend splitting them for easier review and merge.

### 1. Redis `REDIS_KEY_PREFIX` — ✅ Direction is great
**Issue: #199**

The `getDefaultRedisConfig()` change is clean and backward-compatible. One gap: `packages/api/src/scripts/migrate-signals/cli.ts` creates Redis clients via `new Redis(args.redisUrl)` directly, bypassing `createRedisClient()` — so the prefix isolation isn't complete. The issue tracks this.

### 2. ChatContainer thread/store resync — ⚠️ Needs evidence
**Issue: #200**

The logic change looks reasonable but we'd need:
- **Reproduction steps**: What user flow triggers the store/URL desync?
- **A regression test**: To prove the fix and prevent future regressions

### 3. Process cleanup scripts — 🔴 Has blocking issues
**Issue: #201**

Three P1-level problems found (see inline comments):
- **Hardcoded path** `/home/yuhan/cat-cafe-runtime` — won't work for any other user
- **Self-kill risk**: Neither script excludes `$$`/`$PPID`, so running from the project directory can kill the script's own terminal session
- **Linux-only**: `/proc/` filesystem scanning doesn't work on macOS

### 4. Script chmod — ✅ No issue needed
The `download-source-overrides.sh` and `install.sh` permission fixes are fine.

## Suggested Next Steps

1. Split into 2-3 separate PRs, each referencing its issue
2. Redis PR (#199): add the `migrate-signals` coverage → **likely to be accepted quickly**
3. ChatContainer PR (#200): add repro steps + test
4. Kill scripts PR (#201): address the three P1s, or consider if this belongs in the repo vs. personal tooling

Happy to help iterate on any of these! 🐾

---
_Review by 金渐层/Opus-4.6 🐾 + 砚砚/GPT-5.4 🐾_

zts212653:
Closing the original PR as it's been superseded by your split PRs — great job on the quick turnaround! 🎉

- Redis isolation → #202
- ChatContainer fix → #203
- Process cleanup scripts → #204 (closed, upstream solution incoming)

[金渐层/Opus-4.6🐾]

PR #259, opened 2026-03-26
feat(desktop): Tauri v2 macOS desktop app for Cat Café

bouillipx:
## Summary
- Add Tauri v2 macOS desktop shell with embedded Node.js sidecars (API port 13004, Web port 13003)
- Rust-side server lifecycle management: spawn, health check (TCP), WebView navigation, graceful shutdown
- Full build pipeline (`build-distributable.sh`): esbuild API bundle + Next.js standalone + pnpm symlink hoisting + native addon packaging
- Fix `NEXT_PUBLIC_API_URL` build-time inlining so frontend correctly targets port 13004 (not 3004)
- Inherit proxy settings from user shell profile for GUI-launched subprocesses (Claude CLI auth)
- Also includes F132 DARE Out-of-the-Box kickoff doc

## Key files
- `packages/desktop/src-tauri/src/lib.rs` — Rust entry point with sidecar spawning
- `packages/desktop/scripts/build-distributable.sh` — Build pipeline
- `packages/desktop/staging/launch-{api,web}.sh` — Sidecar launchers
- `packages/web/next.config.js` — Conditional `output: 'standalone'`

## Test plan
- [x] Build .dmg via `build-distributable.sh` + `npx tauri build`
- [x] Install to /Applications, launch — both servers start, WebView navigates
- [x] Directory browser, audit events, Socket.IO all functional
- [x] Claude CLI authenticates (proxy/env inheritance working)
- [ ] Test on clean macOS install (no dev tools)


zts212653:
Thanks for the substantial exploration here — this is clearly a serious piece of work, and it also surfaces a real product question for us.

We reviewed this internally from a maintainer / product-direction perspective.

**Current decision: we are not merging this PR into `clowder-ai main` in its current form.**

The main reasons are:

1. **No accepted issue / design anchor yet for this direction.** Desktop packaging is a meaningful product decision for the public repo, and we do not want to merge a large implementation before the direction is explicitly accepted.
2. **It changes the repo’s build/runtime surface too much for `main` right now.** Introducing a desktop toolchain / Rust / packaged desktop workflow into the mainline workspace is not something we can take as an implementation-first change.
3. **The architecture is not the direction we want.** The shell is taking ownership of process orchestration (Tauri/Rust spawning and managing Node sidecars). For our product, we want the opposite boundary: the runtime should own its own lifecycle, and any desktop shell should stay thin.
4. **The PR scope is mixed.** It includes unrelated doc changes, including the DARE/F132 doc conflict, so it is not reviewable as one coherent unit.

What we *do* agree with:

- Desktop use cases are real.
- This exploration is useful and helps expose the constraints we need to think through.

Our current preferred direction is:

- First validate demand with **PWA / Safari Web App**.
- If we later ship a packaged **macOS** client, our current preference is a **thin shell over the existing runtime**, not a shell-managed sidecar architecture.
- If the requirement becomes **cross-platform packaged desktop**, that would need a separate design discussion; it is not something we want to back into via this PR.

So the best next step is **not** to keep iterating on this PR directly. If you want to continue pushing this direction, please open a **smaller discussion / issue focused on desktop direction**, and split any unrelated docs out of this branch. We would be happy to review a scoped proposal from there.

Thanks again for the work and for pushing on an important direction.

— Maintainer review summary by Maine Coon / GPT-5.4

PR #368, opened 2026-04-04
fix: opencode builtin providers get runtime config for unknown models

xu75:
## Problem

金渐层 (opencode) invocations fail with `Model not found: claude-opus-4-6/` because opencode 1.3.7/1.3.13's builtin anthropic provider doesn't recognize newer model IDs.

**Root cause trace:**
1. `cat-config.json` → `defaultModel: "anthropic/claude-opus-4-6"` ✅
2. `invoke-single-cat.ts` → `anthropic` in `BUILTIN_OPENCODE_PROVIDERS` → **skips runtime config generation**
3. opencode CLI receives `-m anthropic/claude-opus-4-6` → builtin model list outdated → ❌ "Model not found"

## Fix

- Remove `BUILTIN_OPENCODE_PROVIDERS` gate — always generate runtime config for opencode cats with `api_key` auth
- Runtime config explicitly registers the model, bypassing opencode's builtin model list
- Add apiType inference from provider name as fallback when `resolvedAccount.protocol` is not set

## Test

Added: `F189: builtin provider (anthropic) also gets runtime config so unknown models are registered`
- All 3 F189 tests pass, no regressions

## Upstream

Filed: anomalyco/opencode#21019

[孟加拉猫/Opus-46🐾]

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,10 +13,17 @@
 # These ports must not conflict with other local services.
 # Default: Frontend 3003, API 3004, Redis 6399
 # Convention: API port = Frontend port + 1 (same as internal 3001→3002)
+#
+# Reverse proxy / remote access 反向代理 / 远端访问:
+#   Behind Nginx (port 80/443), the frontend auto-detects same-origin
+#   and routes /api/ + /socket.io/ through the proxy — no env var needed.
+#   Only set NEXT_PUBLIC_API_URL if you need a non-standard API endpoint.
+#   Also set FRONTEND_URL on the API side for CORS (see below).
 
 FRONTEND_PORT=3003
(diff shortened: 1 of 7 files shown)


xu75:
Closing: all fix commits (`30f6abd`, `164a7f5`) are already on `main` via direct push. The CONFLICTING state is a GitHub artifact — no content remains to merge.

**Verification:**
- `Model not found` issue: fixed ✅ (confirmed by audit logs + test 15/15)
- `apiType` protocol priority P1: tracked separately (see review comment)
- `completed without textual output`: independent issue, not in PR scope

[宪宪/Opus-46🐾]

PR #437, opened 2026-04-11
fix(threads): ZSet index self-heal and startup repair

### LINKED ISSUE #560: Bug: thread list disappears when Redis loses per-user ZSet index after unclean shutdown
## Summary
After an unclean Redis shutdown, the per-user thread list ZSet can lose members while thread detail hashes still survive. When that happens, `/threads` and other list consumers can show only the default thread even though the thread data is still present in Redis.

## Root Cause
We persist thread detail hashes and the per-user thread list in separate Redis structures:
- `thread:{threadId}` hash holds authoritative thread metadata
- `threads:user:{userId}` ZSet is a derived index used by `list()` and startup consumers

If Redis/AOF loses the ZSet members but keeps the hashes, list-side reads stop seeing those threads. Startup evidence rebuild can also misinterpret the missing list entries as stale and delete related indexed state.

## Impact
- Existing threads disappear from the thread list
- Startup consumers that rely on `threadStore.list()` can observe an incomplete thread set
- Evidence index rebuild can run against an incomplete thread list and prune live thread anchors

## Proposed Fix
- Add list-side self-heal for sparse/missing per-user thread indexes
- Add startup repair before evidence index rebuild
- Count repaired users and repaired members explicitly so the repair path is observable

## Related
- PR: #437

08mamba24:
Fixes #560

## Summary
- **list() self-heal**: When `list()` returns ≤1 thread for a user, scans thread Hashes and rebuilds the ZSet index. Gated by 5-min cooldown written only after successful rebuild.
- **repairIndex() startup repair**: Single-pass scan of all thread Hashes on server start, rebuilds sparse ZSet indexes via member diff detection. Runs BEFORE evidence index rebuild to prevent `IndexBuilder` from treating missing threads as stale and deleting them.
- **Single-pass optimization**: `repairIndex()` collects `userId→threadId→lastActiveAt` in one scan and writes missing ZADD entries directly — no re-scan per affected user.

## Root Cause
After an unclean Redis shutdown (e.g. hard power-off), AOF incomplete flush can lose ZSet index data while Hash data survives. This causes "all threads disappeared" for users.

## Test plan
- [x] 28/28 tests pass (5 new: list self-heal, cooldown gate, cross-user isolation, repairIndex sparse rebuild, partial ZSet loss detection)
- [x] Biome format clean
- [ ] Manual verification with Redis FLUSHALL of ZSet only

## Review notes
- P1 fix: `repairIndex()` runs before `indexBuilder.rebuild()` in startup sequence
- P2 fix: true single-pass — Phase 1 collects data, Phase 2 writes directly (no `rebuildUserIndexFromThreadDetails` re-scan)
- Removed unused `rebuildUserIndexForce()` wrapper

### DIFF
--- a/packages/api/src/domains/cats/services/stores/ports/ThreadStore.ts
+++ b/packages/api/src/domains/cats/services/stores/ports/ThreadStore.ts
@@ -278,6 +278,8 @@ export interface IThreadStore {
   restore(threadId: string): boolean | Promise<boolean>;
   /** F095 Phase D: List soft-deleted threads (trash bin). */
   listDeleted(userId: string): Thread[] | Promise<Thread[]>;
+  /** Startup repair: rebuild sparse ZSet indexes from thread detail hashes. */
+  repairIndex?(): Promise<{ repaired: number }>;
 }
 
 const MAX_THREADS = 100;
--- a/packages/api/src/domains/cats/services/stores/ports/ThreadStore.ts
+++ b/packages/api/src/domains/cats/services/stores/ports/ThreadStore.ts
@@ -278,8 +278,8 @@ export interface IThreadStore {
   restore(threadId: string): boolean | Promise<boolean>;
   /** F095 Phase D: List soft-deleted threads (trash bin). */
   listDeleted(userId: string): Thread[] | Promise<Thread[]>;
-  /** Startup repair: rebuild sparse ZSet indexes from thread detail hashes. */
-  repairIndex?(): Promise<{ repaired: number }>;
+  /** Repair sparse/missing per-user thread indexes from authoritative thread detail hashes. */
+  repairIndex?(userId?: string): Promise<{ repairedUsers: number; repairedMembers: number }>;
 }
 
 const MAX_THREADS = 100;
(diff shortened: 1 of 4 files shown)


zts212653:
Maintainer update:
- I pushed `bb5f7094` onto this PR branch to align the repair flow with the source-side fix we validated before intake.
- The branch is now rebased onto current `main` and the PR is linked to accepted issue #560.

Validation I ran locally on this rebased branch:
- `pnpm --filter @cat-cafe/api build` before the rebase passed on the PR branch snapshot
- `REDIS_URL=redis://127.0.0.1:6398/15 CAT_CAFE_REDIS_TEST_ISOLATED=1 node --test packages/api/test/redis-thread-store.test.js` → 33/33 passing after the rebase

Note: `pnpm --filter @cat-cafe/api lint` on top of current `origin/main` currently exposes unrelated upstream baseline type errors outside this PR's 4-file diff, so I used the targeted Redis regression suite as the merge gate for this patch.

— 缅因猫-gpt5.4

Issue #470, opened 2026-04-13
fix(api): allow google clientId with api_key for third-party providers


QiliangLi:
## Problem
`@gemini` mentions were failing with "failed to resolve bound account 'gemini-thirdparty'" error when using third-party API providers like code.newcli.com.

## Root Cause
The validation in `validateRuntimeProviderBinding()` was blocking ALL `api_key` accounts for `google` clientId:

```typescript
// packages/api/src/config/account-resolver.ts:259-261
if (clientId === 'google' && profile.kind !== 'builtin') {
  return 'client "google" only supports builtin Gemini auth';
}
```

But third-party Gemini providers use `api_key` auth with a custom `baseUrl`, which should be allowed.

## Proposed Fix
```diff
-  if (clientId === 'google' && profile.kind !== 'builtin') {
-    return 'client "google" only supports builtin Gemini auth';
+  // Allow api_key accounts for google when using a third-party baseUrl
+  if (clientId === 'google' && profile.kind !== 'builtin' && !profile.baseUrl) {
+    return 'client "google" only supports builtin Gemini auth (or third-party with baseUrl)';
   }
```

## Impact
- ✅ `@gemini` with third-party providers now works
- ✅ Direct Google Gemini API (builtin oauth) still protected
- ✅ Minimal change, low risk

## Commit
Commit hash: \`8c4b7613\` on branch \`runtime/main-sync\`

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

zts212653:
Maintainer triage:

Accepted as an `F127` follow-up enhancement.

What we are accepting in this issue:
- supporting true third-party Gemini gateways for the `google` runtime path
- keeping official Google-owned endpoints on the managed/official auth path

What we are explicitly *not* collapsing into this issue:
- Vertex AI / other official Google Cloud integrations. Those should be handled as their own supported path, not smuggled through the third-party gateway exception.

So this issue is now classified as:
- `enhancement`
- `feature:F127`
- `triaged`
- `accepted`

For PR `#471`, the direction matches this issue, but merge still depends on the PR clearing the remaining contract/test/UI alignment and CI gates.

缅因猫-gpt5.4

mindfn:
[...]

PR #471, opened 2026-04-13
fix(api): allow google clientId with api_key for third-party providers

QiliangLi:
## Summary
- Fixed validation that incorrectly rejected all `api_key` accounts for `google` clientId
- Now allows third-party Gemini providers (with custom `baseUrl`) to work

## Problem
`@gemini` mentions were failing with "failed to resolve bound account 'gemini-thirdparty'" error when using third-party API providers like code.newcli.com.

## Root Cause
The validation in `validateRuntimeProviderBinding()` was blocking ALL `api_key` accounts for `google` clientId:

```typescript
// packages/api/src/config/account-resolver.ts:259-261
if (clientId === 'google' && profile.kind !== 'builtin') {
  return 'client "google" only supports builtin Gemini auth';
}
```

But third-party Gemini providers use `api_key` auth with a custom `baseUrl`, which should be allowed.

## Solution
```diff
-  if (clientId === 'google' && profile.kind !== 'builtin') {
-    return 'client "google" only supports builtin Gemini auth';
+  // Allow api_key accounts for google when using a third-party baseUrl
+  if (clientId === 'google' && profile.kind !== 'builtin' && !profile.baseUrl) {
+    return 'client "google" only supports builtin Gemini auth (or third-party with baseUrl)';
   }
```

## Impact
- ✅ `@gemini` with third-party providers now works
- ✅ Direct Google Gemini API (builtin oauth) still protected
- ✅ Minimal change, low risk

## Test Plan
- [x] `@gemini` with third-party provider works
- [x] Account resolution tests pass
- [x] Built-in oauth protection still works


QiliangLi:
**✅ Fixed in commit 3acd82ba**

Addressed the P2 review feedback about restricting google api_key fallback:
- Added `googleDomains` denylist (generativelanguage.googleapis.com, googleapis.com)  
- Rejects api_key targeting official Google endpoints
- Preserves OAuth-only protection for direct Google access  
- Only true third-party gateways are now allowed

@codex thanks for the catch! 👍

zts212653:
Thanks for chasing this. We reviewed it from two maintainer angles and agree the underlying problem is real: third-party Gemini gateways should not be blocked just because the runtime client is `google`.

That said, we do not want to merge this exact patch as-is yet.

Main reasons:
1. `clowder-ai#470` is still open but not triaged/accepted yet, so our inbound merge gate is not cleared.
2. This change flips an existing contract without updating the surrounding contract surface. On the current PR head, `packages/api/test/cats-routes-runtime-crud.test.js` still has `POST /api/cats rejects non-builtin provider bindings for google client`, and that scenario now returns `201` instead of `400`.
3. The UI/editor side still assumes `google` only exposes builtin accounts (`packages/web/src/components/hub-cat-editor.model.ts` filters Gemini down to builtin profiles only), so backend and frontend are out of sync.
4. Architecturally, this is mixing two different things under one flag: official Google endpoints vs third-party gateways serving Gemini-family models. We should either update that contract end-to-end, or model them separately instead of growing one more special-case in `validateRuntimeProviderBinding()`.

So our current maintainer verdict is: the direction is welcome, but this PR is not merge-ready yet.

The smallest follow-up that would make this much easier to accept:
- triage/accept `#470`
- rebase to latest `main`
- update/add tests for both:
  - allow `api_key + baseUrl` for true third-party gateways
  - still reject Google-owned official endpoints
- align the Hub account filtering/editor behavior with the new backend contract
- keep malformed `baseUrl` handling defensive

Separate note: Vertex AI / official Google endpoints should likely be handled as their own supported path rather than folded into this third-party gateway exception.

If you want, feel free to keep iterating on this PR with that shape. The problem is valid; we just want the contract, tests, and UI assumptions to move together.

— Maintainer synthesis  
宪宪/Opus-46 + 砚砚/GPT-5.4

zts212653:
Hi @QiliangLi — we noticed the branch was updated (merge from `main`), but the substantive blockers from our [previous review](https://github.com/zts212653/clowder-ai/pull/471#issuecomment-4251688761) are still open:

1. **Contract test not synced** — `cats-routes-runtime-crud.test.js` still asserts that `google + third-party api_key` must return `400`. On the current PR head, this test fails (`201 !== 400`). The test needs to be updated to match the new behavior, and a new regression test should be added for the Google-domain denylist path.

2. **Frontend account filter not updated** — `hub-cat-editor.model.ts` (`filterAccounts()`) still restricts `google` to builtin-only profiles, meaning the UI won't let users select a third-party account even though the API now allows it.

3. **No defensive URL parsing** — `new URL(profile.baseUrl)` can throw on malformed input. A `try/catch` returning a validation error would prevent the validation flow from crashing.

On the positive side: [#470](https://github.com/zts212653/clowder-ai/issues/470) has been triaged and accepted (`enhancement` + `feature:F127` + `triaged` + `accepted`), so the issue-first gate is now clear. ✅

Once these three items are addressed, we're happy to re-review. Let us know if you have questions about any of them — we're glad to help.

— 布偶猫-Opus4.6 🐾 + 缅因猫-GPT5.4 🐾

zts212653:
Pushed a small maintainer fixup onto `runtime/main-sync` to unblock this PR.

What changed:
- allow `google` runtime cats to bind third-party gateway accounts
- keep official Google endpoints (`*.googleapis.com`, `generativelanguage.googleapis.com`) on builtin OAuth only
- return a validation error instead of throwing on malformed `baseUrl`
- align Hub account filtering so Google shows builtin auth + allowed third-party gateways, while hiding official Google `api_key` entries
- add a local fallback for `builtinAccountFamilyForClient` on the current head so the updated Hub flow stays runnable during review

Targeted verification I ran locally on the updated branch:
- `pnpm --filter @cat-cafe/api build`
- `node --test packages/api/test/cats-routes-runtime-crud.test.js --test-name-pattern "allows third-party gateway bindings for google client|rejects official Google endpoints|rejects malformed third-party gateway baseUrl"`
- `pnpm exec vitest run src/components/__tests__/hub-cat-editor.test.tsx -t "allows google to use builtin auth plus third-party gateway accounts only|shows third-party google gateways"`

If you want, we can still split the shared fallback into a separate cleanup later; I kept this patch scoped to making the PR mergeable.

缅因猫-gpt5.4

zts212653:
## Maintainer Review — 布偶猫-Opus4.6 🐾

Reviewed head: `afc5f752` (includes @QiliangLi's original fix + 砚砚-GPT5.4's alignment commits)

### Verdict: **LGTM — ready to merge** ✅

All three blockers from our [previous reviews](https://github.com/zts212653/clowder-ai/pull/471#issuecomment-4265249117) are resolved:

| Blocker | Resolution |
|---------|-----------|
| Contract test not synced | Old reject test → allow test; 2 new regression tests (official Google domain rejection + malformed URL rejection) |
| Frontend account filter locked to builtin-only | `filterAccounts()` now includes valid third-party gateways; `AccountSection` secondary filter removed; warning text updated |
| No defensive URL parsing | `parseHostname()` with try/catch; malformed URLs return a clean validation error |

### Test coverage

**API** (3 targeted tests):
- `allows third-party gateway bindings for google client` — 201
- `rejects official Google endpoints for google api_key bindings` — 400 + message match
- `rejects malformed third-party gateway baseUrl for google client` — 400 + message match

**Web** (2 targeted tests):
- `allows google to use builtin auth plus third-party gateway accounts only` — filterProfiles returns [builtin, third-party] only
- `shows third-party google gateways in the account selector` — dropdown includes proxy, excludes official Google API key

### Minor note (non-blocking)

`GOOGLE_OWNED_DOMAINS` / `parseHostname()` / `isOfficialGoogleHostname()` are duplicated between API and Web. We'll consolidate into `@cat-cafe/shared` during our internal intake. Does not block this merge.

Thanks @QiliangLi for the original fix and for opening this path — the problem you identified is real and we're glad to ship it. 🎉

— 布偶猫-Opus4.6 🐾

PR #482, opened 2026-04-14
fix(antigravity): use getComputedStyle for visibility, regex icon cleanup

kaltsonikostaki848-crypto:
## Summary
Addresses code review feedback from @codex on commit 45483082.

### Changes
**P1 - Visibility detection:**
- Replace class-name matching with `isVisiblyHidden()` helper using `getComputedStyle` + `clientHeight` + `aria-hidden` + `inert`
- Handles edge case where `opacity-0` class is overridden by state classes

**P1 - Icon text leak:**
- Add `ICON_CONCAT_RE` regex to strip concatenated icon name runs
- Expand icon name list (+20 entries)

**Bug fixes discovered during testing:**
- `extractBlockText` root guard blocking thinking extraction; added `skipRootGuard` parameter
- Test fixture DOM `.group` → `.group.pt-4`

### Verification
- 13/13 tests pass ✅
- TypeScript: zero errors ✅

### DIFF
--- a/packages/api/src/domains/cats/services/agents/providers/antigravity/cdp-dom-scripts.ts
+++ b/packages/api/src/domains/cats/services/agents/providers/antigravity/cdp-dom-scripts.ts
@@ -80,7 +80,7 @@ export const POLL_RESPONSE_JS = `(() => {
       }
     }
     // Strip hidden subtrees, icons, scripts, styles
-    clone.querySelectorAll('style, script, [aria-hidden="true"], .google-symbols, [class*="symbol"], [class*="material-symbols"]').forEach((el) => el.remove());
+    clone.querySelectorAll('style, script, [aria-hidden="true"], [inert], .google-symbols, [class*="symbol"], [class*="material-symbols"]').forEach((el) => el.remove());
     for (const el of clone.querySelectorAll('*')) {
       const cls = el.className || '';
       if (typeof cls === 'string' && (/\\bmax-h-0\\b/.test(cls) || /\\bopacity-0\\b/.test(cls) || /\\bhidden\\b/.test(cls))) {
--- a/packages/api/src/domains/cats/services/agents/providers/antigravity/cdp-dom-scripts.ts
+++ b/packages/api/src/domains/cats/services/agents/providers/antigravity/cdp-dom-scripts.ts
@@ -23,6 +23,27 @@
  *                  └ ...
  */
 export const POLL_RESPONSE_JS = `(() => {
+  // --- 0. Visibility helpers ---
+  // Check if an element is visually hidden using computed styles (reliable),
+  // not class names alone (fragile — e.g. opacity-0 can be overridden by state classes).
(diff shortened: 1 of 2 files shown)


zts212653:
Thanks for the thorough work here — the `getComputedStyle` approach is genuinely more robust than class-name matching, and the icon-text cleanup regex is a clever catch. The test fixtures are solid too.

However, I want to be transparent about where this codebase is heading architecturally, because it directly affects whether this PR has a path to merge.

## The CDP DOM scraping approach is being replaced

We've been running the CDP DOM hack in production since March and hit a pattern: **every Antigravity IDE update breaks the scraper** — class names change, DOM structure shifts, new element types appear. The fixes are always the same shape: more heuristics, more selectors, more edge-case guards. This PR is a good example of that cycle.

The fundamental problem isn't *how* we scrape the DOM — it's *that* we scrape the DOM at all.

## What we're doing instead: ConnectRPC Bridge

We've already shipped (internally) a complete replacement that talks directly to Antigravity's Language Server via ConnectRPC/gRPC. The architecture:

```
Cat Cafe AgentRouter
  → AntigravityAgentService
    → AntigravityBridge (ConnectRPC transport)
      → Antigravity LS (gRPC, structured TrajectoryStep data)
```

Every problem this PR fixes **doesn't exist** in the Bridge approach:

| DOM scraping problem | This PR's fix | Bridge: why it's a non-issue |
|---|---|---|
| Icon text leaks into response | 40-name regex + `ICON_CONCAT_RE` | Data arrives as structured `TrajectoryStep` objects — no DOM, no icons |
| Visibility detection fragile | `getComputedStyle` + `aria-hidden` + `inert` | No DOM involved, no visibility concept |
| User message mis-detection | `.group.pt-4` filtering + system prompt guards | Step type enum: `CORTEX_STEP_TYPE_USER_INPUT` |
| Assistant block location | 6-level ancestor walk | `CORTEX_STEP_TYPE_PLANNER_RESPONSE` — direct field access |
| Thinking separation | CSS class guessing | `plannerResponse.thinking` field |

The Bridge code is ~180 lines of typed TypeScript with a 6-bucket step classifier, vs the ~275-line inline JS string that has to guess at DOM structure.

## What this means for this PR

We plan to sync the Bridge architecture to this repo, which will **remove** `cdp-dom-scripts.ts` and `AntigravityCdpClient.ts` entirely. Merging this PR would create conflicts with that sync and send the wrong signal about the maintenance direction of the CDP path.

## What we'd welcome instead

The failure modes you identified (icon text leak, visibility false positives, thinking extraction bugs) are valuable as **regression test cases** for the Bridge architecture. If you're interested in contributing to the active codebase, the Bridge layer has open work:

- **Streaming activity signals** — heartbeat during long tool execution so the frontend knows the model is still working
- **Stall error deduplication** — the same error sometimes surfaces twice in the frontend

We appreciate the contribution and the clear problem identification. Just wanted to explain the architectural context so you understand the decision.

PR #489, opened 2026-04-15
feat(F153): inline @mention detection observability — counters + shadow detection

bouillipx:
## Summary

- 8+1 OTel counters for inline action @mention detection pipeline (checked/detected/shadow_miss/feedback_written/feedback_write_failed/hint_emitted/hint_emit_failed/routed_set_skip + line_start baseline)
- In-process shadow detection: relaxed regex pass diffs with strict detection to find vocab gap candidates
- Shadow miss metadata is hash+length only (no raw text), per data minimization feedback from mindfn (#479)
- Silent `catch {}` blocks now increment failure counters, making pipeline faults visible

## Test plan

- [x] 12/12 new mention-observability tests pass
- [x] 68/68 existing a2a-mentions tests pass (regression)
- [x] 82/82 route-strategies tests pass (regression)
- [x] `pnpm check` lint clean
- [x] All counter names follow `cat_cafe.a2a.*` prefix, attributes only use `agent.id` (F152 metric allowlist compliant)

Refs: #479

🐾 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/api/src/domains/cats/services/agents/routing/a2a-mentions.ts
+++ b/packages/api/src/domains/cats/services/agents/routing/a2a-mentions.ts
@@ -276,9 +276,10 @@ export function detectInlineActionMentionsWithShadow(
   if (!text) return { strictHits: [], shadowMisses: [], routedSetSkips: 0 };
 
   const strictHits = detectInlineActionMentions(text, currentCatId, routedMentions);
-  const strictCatIds = new Set(strictHits.map((h) => h.catId));
 
-  // Relaxed pass: find ALL inline @mentions (no action keyword required)
+  // Relaxed pass: find ALL inline @mentions, classify per occurrence.
+  // P2-1 fix: check action keywords per occurrence instead of global catId diff.
+  // P2-2 fix: only count routedSetSkip when action keyword present, dedup per (catId, line).
   const stripped = text.replace(/```[\s\S]*?```/g, '');
   const allConfigs = Object.keys(catRegistry.getAllConfigs()).length > 0 ? catRegistry.getAllConfigs() : CAT_CONFIGS;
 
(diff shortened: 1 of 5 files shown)


zts212653:
Thanks for the update. I can see `#489` has been updated today, and `#479` is now triaged / anchored under F153.

From the maintainer side, the **problem definition is accepted**. The remaining point I want us to converge on is code shape:
- keep `a2a-mentions` as the owner of mention grammar
- avoid long-term coupling from exporting internal regex just to support shadow detection
- keep `route-serial` from growing further if the observability handling can be extracted cleanly

I'm re-evaluating the latest head with that in mind. If we can keep the capability while tightening the boundary, this is in a mergeable direction.

— 缅因猫-gpt5.4

zts212653:
## Maintainer Review — Both reviewers approve ✅

We've reviewed the latest 4 commits (up to `9386373`) in detail. Both maintainer reviewers now vote **YES to merge**.

### What changed our mind (from initial "hold" to "approve")

The 3 rounds of iteration (P2-1 through P2-4) addressed every correctness concern we raised:

- **P2-1/P2-2**: Shadow and routedSetSkip counting now uses per-occurrence semantics with action-keyword gating — narrative mentions no longer inflate metrics
- **P2-3**: The `RELAXED_*_ACTION_RE` gate creates a clean three-tier classification (strict hit / shadow miss / narrative ignore), which is the right model
- **P2-4**: Same-line dual mention scanning correctly handles narrative-then-actionable sequences

### Remaining architectural preferences (non-blocking)

We initially had concerns about regex exports and double-pass scanning. On re-review:
- The exported constants serve as the strict baseline for shadow diffing — this dependency relationship is correct
- Separation of concerns (strict → routing, relaxed → observability) outweighs single-pass purity
- Counter granularity replaces silent `catch {}` blocks with observable signals — net positive

These are "we might refactor later" items, not merge blockers.

### Next steps

This is a Feature-class PR (`feature:F153`), so final merge requires CVO approval per our governance. We're escalating now.

Intake plan: `absorbed` via Strategy A (merge → full sync covers).

— 宪宪/Opus-46🐾 + 砚砚/GPT-5.4🐾

PR #529, opened 2026-04-19
fix(web): per-slot cancel and correct active cat display

### LINKED ISSUE #534: Stop/cancel and active-cat display drift out of sync in multi-cat web UI
## Problem

Three web-side bugs are currently coupled in multi-cat execution:

1. `targetCats` can remain stale after the last background invocation completes, so the wrong cat keeps showing in the UI.
2. Stop can degrade to thread-level `cancelAll` instead of per-slot cancel when exactly one cat is active.
3. `ThinkingIndicator` and cancel controls can still disappear even when there is exactly one real active slot, because display logic falls back too early to stale or empty `targetCats`.

## Expected

The visible active cat and the cancel target should come from the same authoritative slot truth when available, and stale `targetCats` should not keep old cats visible after the last slot ends.

## Reproduction signal

Maintainer live repro in Cat Cafe / Clowder-derived UI: assistant is visibly streaming but the cancel affordance is missing or targets the wrong cat.

## Candidate fix

- clear `targetCats` when tracked background slots transition from >0 to 0
- emit per-slot cancel when exactly one slot is active
- make `ThinkingIndicator` prefer a single authoritative slot even when `targetCats` is empty or stale-multi

## Linked PR

- Proposed fix: #529

08mamba24:
## Summary

- **P1 — targetCats 状态污染**：background thread 最后一个 invocation 完成后，`markThreadInvocationComplete` 不清理 `targetCats`，导致旧猫残留、前端显示错猫。修复：`slotsAfter === 0` 时调用 `replaceThreadTargetCats(threadId, [])`
- **P1 — Stop 全线程 cancelAll**：`cancelInvocation` 不带 `catId`，服务端走 `cancelAll(threadId)` 打断所有猫。修复：接受可选 `catId`，单猫时 emit 带 catId；`handleStop` 从目标线程 slots 推导 catId（`getThreadState(threadId)`）
- **P1 — ThinkingIndicator 显示/取消不同源**：优先从 `activeInvocations` 单 slot 推导 catId，fallback `targetCats[0]`，确保 cancel 按钮和显示猫同源
- **P2 — RightStatusPanel/ParallelStatusBar/MobileStatusSheet**：active cat 显示改为 slot-first（`deriveActiveCats` 新函数），`status-helpers` 新增 `collectSnapshotActiveCats`

## Root cause

前端把 route intent（`targetCats`）当 active state 用 + 通用 Stop 是 cancelAll。两个源头独立修复后，串台/误打断问题关闭。

## Test plan

- [x] `useSocket-background.test.ts` 58 pass — background completion 清理、多猫最后一只清理、单猫 partial 不清理
- [x] `status-helpers-liveness.test.ts` 8 pass — slot-first 推导、legacy fallback
- [x] `useAgentMessages-loading.test.ts` — 有 `act()` env 失败（pre-existing，main 也失败）；新增 background stop catId 路由测试逻辑已验证通过
- [x] 无新增 Biome error，无新增 TS error

## Issue Closure

- Closes #534

### DIFF
--- a/packages/web/src/components/ChatContainer.tsx
+++ b/packages/web/src/components/ChatContainer.tsx
@@ -65,6 +65,7 @@ export function ChatContainer({ threadId }: ChatContainerProps) {
   const {
     messages,
     hasActiveInvocation,
+    activeInvocations,
     intentMode,
     targetCats,
     catStatuses,
@@ -768,6 +769,8 @@ export function ChatContainer({ threadId }: ChatContainerProps) {
             targetCats={targetCats}
             catStatuses={catStatuses}
             catInvocations={catInvocations}
+            activeInvocations={activeInvocations}
+            hasActiveInvocation={hasActiveInvocation}
             threadId={threadId}
             messageSummary={messageSummary}
             width={statusPanelWidth}
@@ -787,6 +790,8 @@ export function ChatContainer({ threadId }: ChatContainerProps) {
         targetCats={targetCats}
         catStatuses={catStatuses}
         catInvocations={catInvocations}
+        activeInvocations={activeInvocations}
+        hasActiveInvocation={hasActiveInvocation}
         threadId={threadId}
         messageSummary={messageSummary}
       />
--- a/packages/web/src/components/MobileStatusSheet.tsx
+++ b/packages/web/src/components/MobileStatusSheet.tsx
@@ -2,7 +2,6 @@
 
 import { useMemo } from 'react';
 import { formatCatName, useCatData } from '@/hooks/useCatData';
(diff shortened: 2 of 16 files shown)


zts212653:
I pushed a follow-up fix onto the PR branch: `5a5fd613`.

It closes the remaining ThinkingIndicator hole where a single live slot could still fail to render/cancel if `targetCats` was empty or stale-multi. The component now prefers the single authoritative slot even when `targetCats` is not length 1, and the liveness tests cover both `targetCats=[]` and stale-multi fallback cases.

PR #621, opened 2026-04-30
fix(windows): replace redis-cli dependency with pure-PowerShell RESP PING/PONG (#620)

### LINKED ISSUE #620: Bug: Windows WSL Redis 被 start-windows.ps1 误判为不可用
## What happened?

Windows 环境下 Redis 运行在 WSL 内（端口 6379），`.env` 配置 `REDIS_URL=redis://localhost:6379` 且 `REDIS_PORT=6379`。运行 `pnpm start` 后提示 **"Redis not running on port 6379"**，随后降级为内存模式。实际上 WSL 内的 Redis 服务完全正常，`localhost:6379` 对 Windows TCP 可达。

## Steps to reproduce

1. Windows 11 + WSL2，WSL 内安装并运行 Redis（默认 `127.0.0.1:6379`）
2. `.env` 配置：
   ```
   REDIS_URL=redis://localhost:6379
   REDIS_PORT=6379
   ```
3. 确保 Windows 上没有安装 `redis-cli.exe`（即没有 Windows 原生 Redis 二进制文件）
4. 在项目根目录运行 `pnpm start`
5. 观察输出：`Redis not running on port 6379` → 降级为 memory 模式

## Expected behavior

脚本正确探测到 `localhost:6379` 上的 Redis 服务可达，正常使用 Redis 启动。

## Environment

- OS: Windows 11 + WSL2
- Node: v22.x
- pnpm: 9.x
- Redis: 运行在 WSL 内

pencilkill:
## Summary

- Replace `redis-cli.exe` dependency in `start-windows.ps1` and `stop-windows.ps1` with pure-PowerShell TCP + RESP PING/PONG
- Fixes #620: WSL Redis falsely reported as "not running" when `redis-cli.exe` is unavailable on Windows

### `scripts/install-windows-helpers.ps1`
Add three new functions:
- `Format-RedisRespCommand` — build RESP protocol command strings
- `Test-RedisReachable` — TCP connect + RESP AUTH + PING/PONG validation
- `Send-RedisShutdown` — TCP connect + RESP AUTH + SHUTDOWN SAVE

### `scripts/start-windows.ps1`
Refactor Storage section:
- When `REDIS_URL` is configured: validate connectivity via `Test-RedisReachable`
- When no `REDIS_URL`: check managed port, attempt to start own Redis on fallback
- Remove `Test-LocalRedisUrl` usage in startup path
- Remove all `redis-cli` calls

### `scripts/stop-windows.ps1`
- Replace `redis-cli ping`/`shutdown` with `Test-RedisReachable` + `Send-RedisShutdown`
- Replace `Test-LocalRedisUrl` call with inline loopback check
- Remove `redis-cli` binary resolution

### Test updates
- Update `windows-portable-redis-lifecycle.test.js` and `windows-portable-redis-url.test.js` to match new code patterns

## Test plan

- [ ] `pnpm start` on Windows with WSL Redis (REDIS_URL=redis://localhost:6379)
- [ ] `pnpm start` on Windows with no local Redis (should attempt start or memory fallback)
- [ ] `pnpm start --memory` (should skip Redis entirely)
- [ ] `pnpm stop` on Windows
- [ ] Existing test suite: `node --test packages/api/test/windows-portable-redis-*.test.js`

Closes #620

### DIFF
--- a/packages/api/test/windows-portable-redis-lifecycle.test.js
+++ b/packages/api/test/windows-portable-redis-lifecycle.test.js
@@ -83,14 +83,13 @@ test('Windows stop script only stops Clowder-owned API and frontend listeners', 
 
 test('Windows startup preserves runtime Redis overrides, validates artifacts, and exits when service jobs stop', () => {
   assert.match(startWindowsScript, /\$configuredRedisUrl = if \(\$env:REDIS_URL\)/);
-  assert.match(helpersScript, /function Test-LocalRedisUrl/);
+  assert.match(helpersScript, /function Test-RedisReachable/);
+  assert.match(helpersScript, /function Format-RedisRespCommand/);
+  assert.match(helpersScript, /function Send-RedisShutdown/);
   assert.match(helpersScript, /function Get-RedactedRedisUrl/);
-  assert.match(
-    startWindowsScript,
-    /\$useExternalRedis = \$useRedis -and \$configuredRedisUrl -and -not \(Test-LocalRedisUrl -RedisUrl \$configuredRedisUrl -RedisPort \$RedisPort\)/,
-  );
   assert.match(startWindowsScript, /\$safeConfiguredRedisUrl = Get-RedactedRedisUrl -RedisUrl \$configuredRedisUrl/);
-  assert.match(startWindowsScript, /Write-Ok "Using external Redis: \$safeConfiguredRedisUrl"/);
+  assert.match(startWindowsScript, /Test-RedisReachable -RedisUrl \$configuredRedisUrl/);
+  assert.match(startWindowsScript, /Write-Ok "Redis reachable at \$safeConfiguredRedisUrl"/);
   assert.match(startWindowsScript, /\$safeEffectiveRedisUrl = Get-RedactedRedisUrl -RedisUrl \$effectiveRedisUrl/);
   assert.match(
     startWindowsScript,
@@ -111,11 +110,9 @@ test('Windows startup preserves runtime Redis overrides, validates artifacts, an
   assert.match(startWindowsScript, /Service job '\$\(\$job.Name\)' stopped \(\$\(\$job.State\)\)/);
 });
 
-test('Windows startup preserves configured REDIS_URL with DB suffix and credentials when local Redis is already running', () => {
-  assert.match(
-    startWindowsScript,
-    /if \(\$configuredRedisUrl\) \{\s+\$env:REDIS_URL = \$configuredRedisUrl\s+\} else \{\s+\$env:REDIS_URL = "redis:\/\/localhost:\$RedisPort"\s+\}/s,
-  );
+test('Windows startup preserves configured REDIS_URL when provided', () => {
+  assert.match(startWindowsScript, /\$env:REDIS_URL = \$configuredRedisUrl/);
+  assert.match(startWindowsScript, /Write-Ok "Redis reachable at \$safeConfiguredRedisUrl"/);
 });
 
 test('Windows startup refuses non-Clowder Redis listeners before reusing port 6399', () => {
@@ -192,13 +189,11 @@ test('Windows PATH refresh preserves shell-provided shim entries while appending
   );
 });
 
-test('Windows stop script resolves redis-cli through the shared helper chain before shutdown', () => {
+test('Windows stop script resolves Redis through RESP instead of redis-cli', () => {
   assert.match(stopWindowsScript, /install-windows-helpers\.ps1/);
-  assert.match(stopWindowsScript, /Resolve-PortableRedisBinaries -ProjectRoot \$ProjectRoot/);
   assert.match(stopWindowsScript, /Resolve-PortableRedisLayout -ProjectRoot \$ProjectRoot/);
-  assert.match(stopWindowsScript, /Resolve-GlobalRedisBinaries/);
-  assert.match(stopWindowsScript, /\$redisCli = \$redisCommands\.CliPath/);
-  assert.doesNotMatch(stopWindowsScript, /& redis-cli -p \$RedisPort ping/);
+  assert.match(stopWindowsScript, /Test-RedisReachable -RedisUrl \$shutdownUrl/);
+  assert.match(stopWindowsScript, /Send-RedisShutdown -RedisUrl \$shutdownUrl/);
   assert.match(
     stopWindowsScript,
     /\$redisPidFile = if \(\$redisLayout\) \{ Join-Path \$redisLayout\.Data "redis-\$RedisPort\.pid" \} else \{ \$null \}/,
@@ -213,9 +208,8 @@ test('Windows stop script resolves redis-cli through the shared helper chain bef
     /\$isClowderOwned = \$isManagedPid -or \(Test-ClowderOwnedProcess -ProcessId \$conn\.OwningProcess -ClowderProjectRoot \$ProjectRoot\)/,
   );
   assert.match(stopWindowsScript, /Write-Warn "Skipping non-Clowder Redis listener on port \$RedisPort/);
-  assert.match(stopWindowsScript, /Get-RedisAuthArgs\s+-RedisUrl\s+\$configuredRedisUrl/);
-  assert.match(stopWindowsScript, /@redisAuthArgs\s+ping/);
-  assert.match(stopWindowsScript, /@redisAuthArgs\s+shutdown/);
+  assert.doesNotMatch(stopWindowsScript, /& redis-cli/);
+  assert.doesNotMatch(stopWindowsScript, /@redisAuthArgs/);
 });
 
 test('Windows start.bat delegates to start-windows.ps1', () => {
--- a/packages/api/test/windows-portable-redis-url.test.js
+++ b/packages/api/test/windows-portable-redis-url.test.js
@@ -106,6 +106,12 @@ test('Windows Redis auth helpers decode percent-escaped ACL credentials before i
   );
 });
 
+test('Windows RESP helpers use UTF-8 byte counts for bulk string lengths', () => {
+  assert.match(helpersScript, /function Format-RedisRespCommand/);
+  assert.match(helpersScript, /\[System\.Text\.Encoding\]::UTF8\.GetByteCount\(\$arg\)/);
+  assert.doesNotMatch(helpersScript, /\$\(\$arg\.Length\)/);
+});
+
 test('Windows portable Redis defers REDIS_URL to runtime instead of hardcoding localhost:6379', () => {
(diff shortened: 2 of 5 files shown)


mindfn:
@codex review

mindfn:
Thanks for the contribution here. Removing the Windows-side `redis-cli.exe` dependency is the right direction, and the WSL Redis false-negative in #620 is worth fixing. I am requesting changes because the current implementation drops existing managed/local Redis safety semantics.

Review result: requesting changes.

P1 blocking: the declared Windows Redis test slice fails on this branch. I ran:

```bash
node --test packages/api/test/windows-portable-redis-*.test.js
```

Result: 56 pass, 2 fail:
- `Windows PowerShell scripts stay ASCII-only to avoid console codepage issues`
- `Windows startup preserves configured REDIS_URL with DB suffix after Redis auto-start`

The second failure is the real behavior regression. `start-windows.ps1` now treats any configured `REDIS_URL` as user-managed: it only probes it, and if it is down it falls back to memory storage (`scripts/start-windows.ps1`, current lines 311-319). That breaks the existing local-managed case where `REDIS_URL` is a same-port loopback URL with auth or DB suffix, e.g. `redis://:pw@localhost:6399/15`: after reboot, the script no longer starts the portable Redis bundle and no longer preserves the exact configured URL. Main previously started Redis and then re-exported `$configuredRedisUrl` in both already-running and auto-start branches. Please keep the WSL/external case, but still classify same-port loopback URLs as managed: start portable Redis with the configured auth args and set `$env:REDIS_URL = $configuredRedisUrl` after startup succeeds.

P1 blocking: `stop-windows.ps1` can now send `SHUTDOWN SAVE` to a loopback Redis port that it did not prove belongs to Clowder. The new external check only tests whether the configured host is loopback (`scripts/stop-windows.ps1`, current lines 132-142); it no longer verifies that the configured URL port matches `$RedisPort`. Later it proves ownership only for listeners on `$RedisPort`, but sends shutdown to `$configuredRedisUrl` (`scripts/stop-windows.ps1`, current lines 151-171). With `REDIS_URL=redis://localhost:6379` plus a Clowder-owned listener on 6399, the script can shut down 6379. The old `Test-LocalRedisUrl` explicitly rejected port mismatches. Please restore the port check, or send shutdown only to the verified managed port.

P2: `Format-RedisRespCommand` uses `$arg.Length` for RESP bulk lengths (`scripts/install-windows-helpers.ps1`, current lines 262-268), but the stream writer emits UTF-8. RESP bulk lengths are byte counts, not .NET character counts. Percent-decoded non-ASCII Redis credentials will produce malformed AUTH packets. Use UTF-8 byte count (`[System.Text.Encoding]::UTF8.GetByteCount($arg)`) or write the RESP command as raw bytes from the same encoding.

P3: the new comments in `start-windows.ps1` contain em dashes, which violates the existing ASCII-only Windows script contract and causes the first failing test above. Easy cleanup once the blocking Redis behavior is fixed.

[砚砚/GPT-5.5🐾]

mindfn (dismissed):
Thanks for the contribution. Requesting changes; see detailed review comment: https://github.com/zts212653/clowder-ai/pull/621#issuecomment-4351023429

[砚砚/GPT-5.5🐾]

mindfn:
Thanks for pushing this Windows Redis fix. Removing the `redis-cli.exe` requirement is valuable for WSL Redis users; the comments below are about preserving the existing managed/local Redis safety contracts while making that improvement.

## Supplementary review — confirming and extending @mindfn's findings

I independently reviewed the diff and confirm all four findings from the previous review. Adding additional context:

### P1 confirmations with precision

**P1-1 (`start-windows.ps1` L311-319)**: The root cause is that the old `Test-LocalRedisUrl` classification (loopback + port match → managed) is completely removed. The new code treats **any** `$configuredRedisUrl` as user-managed and only probes connectivity. This means `REDIS_URL=redis://:pw@localhost:6399/15` (loopback URL on the managed port with auth + DB suffix) will no longer trigger managed Redis startup. When Redis is down after a reboot, the script silently degrades to memory instead of starting its own Redis.

**P1-2 (`stop-windows.ps1` L132-171)**: To be precise about the attack vector: with `REDIS_URL=redis://localhost:6379` and `$RedisPort=6399`:
1. `$isExternalRedis = $false` (localhost is loopback ✓)
2. Ownership check passes on port **6399** (finds Clowder-owned listener ✓)
3. `$shutdownUrl = $configuredRedisUrl` → `redis://localhost:6379`
4. `Send-RedisShutdown` sends `SHUTDOWN SAVE` to port **6379** — a Redis Clowder doesn't own

The missing ingredient is the port comparison that `Test-LocalRedisUrl` previously enforced. The fix is straightforward: add `$uri.Port -ne $RedisPort` to the external-host check (or equivalently, always send shutdown to `redis://localhost:$RedisPort` when we've verified ownership on that port).

### P2 confirmation (`Format-RedisRespCommand`)

`$arg.Length` returns .NET character count, but the `StreamWriter` emits UTF-8 bytes. RESP bulk length must match byte count. For pure ASCII credentials this happens to be identical, but percent-decoded non-ASCII chars (e.g. `redis://:p%C3%A4ss@...`) will produce a malformed RESP frame. Use `[System.Text.Encoding]::UTF8.GetByteCount($arg)`.

### P3 — supplementary: code duplication

`Test-RedisReachable` and `Send-RedisShutdown` share ~40 lines of identical connection + AUTH logic. Consider extracting the TCP connect + RESP AUTH sequence into a shared helper, reducing both functions to "connect → do one thing → close."

### CI note

All 5 CI checks pass because the PR updated test assertions to match the new (regressed) behavior. The two locally failing tests (`ASCII-only` + `preserves configured REDIS_URL with DB suffix`) correctly catch the regressions but aren't surfaced in CI because the assertions were weakened. Reviewers should not rely on green CI alone here.

---

**Conclusion**: Agree with CHANGES_REQUESTED. The two P1s are blocking; both are regressions from removing `Test-LocalRedisUrl` without preserving its port-matching semantics. The RESP byte-count P2 should also be fixed before merge.

[宪宪/Opus-46🐾]

zts212653:
Maintainer synthesis after the two reviews:

This PR is solving a real Windows/WSL Redis problem, and the direction of removing the Windows-side `redis-cli.exe` dependency is good. The part we cannot accept yet is the broader rewrite of the managed-vs-external Redis control flow. The fix should separate two concerns that are currently coupled in this branch:

1. **Reachability:** can this Redis endpoint answer RESP `PING` without requiring `redis-cli.exe`?
2. **Ownership / process management:** is this Redis instance one that Clowder is allowed to start or stop?

That distinction gives us the shape I would be comfortable merging:

- A configured `REDIS_URL` that is already reachable via RESP should be usable as-is. This preserves the WSL Redis case that motivated #620.
- A configured same-port loopback URL that is **not** reachable should still be eligible for managed portable Redis startup. After startup, keep the exact configured URL, including auth and DB suffix, instead of replacing it with `redis://localhost:$RedisPort`.
- Shutdown must only target the Redis endpoint whose ownership was verified. If the configured URL points to a different port than `$RedisPort`, treat it as external or send shutdown only to the verified managed port. Never verify ownership on one port and send `SHUTDOWN SAVE` to another.
- The RESP command builder needs byte lengths, not character lengths. Use UTF-8 byte counts or write raw UTF-8 bytes consistently.
- Please keep the Windows PowerShell scripts ASCII-only.
- Please add/restore tests for the important behavior contracts, especially:
  - WSL/external Redis reachable without Windows `redis-cli.exe`
  - same-port loopback `REDIS_URL` with auth/DB suffix auto-starts managed Redis when down and preserves the exact URL
  - stop does not shut down a different loopback Redis port
  - non-ASCII Redis credentials produce valid RESP AUTH frames

So the maintainer position is: the problem and transport-layer idea are welcome, but this branch needs to preserve the existing managed Redis safety semantics before merge. If the rework becomes too large, we can also take over with a smaller maintainer patch that keeps the RESP probe while minimizing control-flow changes.

[砚砚/GPT-5.5🐾]

zts212653:
Pushed a maintainer fix commit to this branch: dc5d09080.\n\nWhat changed:\n- keep reachable configured Redis URLs working without Windows redis-cli.exe\n- preserve managed same-port loopback startup and exact REDIS_URL after auto-start\n- restore shutdown safety by using the existing Test-LocalRedisUrl port check\n- use UTF-8 byte counts for RESP bulk strings\n- add tests for the restored contracts\n\nLocal verification:\n- node --test packages/api/test/windows-portable-redis-*.test.js -> 59 pass / 0 fail\n- pnpm check -> pass\n- pnpm lint -> pass, with existing hardcoded-color warnings unrelated to this PR\n\n[砚砚/GPT-5.5🐾]

zts212653 (approved):
## Review: dc5d09080 — maintainer fix

Independent cross-individual review (砚砚 authored fix, 宪宪 reviews).

### Verdict: APPROVE

砚砚的 fix commit 修掉了 P1/P2/P3，同时保留了原 PR 的 RESP 方向。逐项确认：

**P1-1 managed Redis auto-start** ✓
`$configuredIsManagedRedis` 重新用 `Test-LocalRedisUrl`（loopback + port match）做分类。`REDIS_URL=redis://:pw@localhost:6399/15` 在 Redis down 时仍走 managed path → auto-start → probe with configured URL → preserve full URL。

**P1-2 stop port mismatch** ✓
`stop-windows.ps1` 恢复 `Test-LocalRedisUrl -RedisUrl $configuredRedisUrl -RedisPort $RedisPort`。`REDIS_URL=redis://localhost:6379` + `$RedisPort=6399` → port mismatch → skip shutdown。`$shutdownUrl` 只在 `Test-LocalRedisUrl` 通过后才执行，port 已验证。

**P2 RESP byte count** ✓
`Format-RedisRespCommand` 改用 `[System.Text.Encoding]::UTF8.GetByteCount($arg)`。新增专项测试断言 `doesNotMatch` 旧的 `$arg.Length`。

**P3 ASCII-only** ✓
注释中的 em dash 已替换为 semicolons。

**控制流总结**

start 三路分支：
1. configured + reachable → 直接用（WSL fix 命中这条）
2. configured + external + not reachable → memory fallback
3. managed path → auto-start + ownership check + preserve configured URL

stop 入口：`Test-LocalRedisUrl` gate → only shutdown verified managed port。

**测试**: 59/59，覆盖 RESP byte count、`Test-LocalRedisUrl` 在 start/stop 双向出现、managed URL preservation。

**Minor（non-blocking）**:
- `Test-RedisReachable` 和 `Send-RedisShutdown` 有 ~40 行重复的 TCP connect + AUTH 逻辑，可抽共享 helper。留给后续。
- `rediss://` TLS 不处理，但原代码也没处理，非回归。

[宪宪/Opus-46🐾]

PR #631, opened 2026-05-01
fix(api): dedupe queue replay by idempotency key

### LINKED ISSUE #630: Queue mode replay can duplicate execution and waste tokens (missing idempotency dedupe)
## Summary

When a user message is replayed (same logical request) while the thread is in queued delivery flow, the API can create duplicate queue rows and duplicate user messages, which can trigger duplicate downstream execution.

This leads to extra token consumption and duplicated work.

## Why this matters

In production, transport/CLI failures and reconnect retries can replay the same logical user message. If replay is not idempotent in queue mode, we can execute the same task twice.

Observed effect:
- duplicate queued entries for one logical request
- duplicate user messages in thread timeline
- duplicated agent execution and token spend

## Root cause (current behavior)

`POST /api/messages` already resolves an `idempotencyKey`, but in queue-related paths:

1. queue enqueue did not persist/use that key for active-entry dedupe
2. queued `messageStore.append(...)` did not pass `idempotencyKey`

So replayed requests in queue mode (or TOCTOU degrade-to-queue path) could append again and enqueue again.

## Reproduction (before fix)

1. Keep a thread busy so `deliveryMode=queue` path is taken.
2. Send a message request, then replay the same logical request (same idempotency key).
3. Observe another queued entry and another appended user message.

## Expected

Replay with same `idempotencyKey` should return the existing active queue entry/message, without creating a new queue row or new user message append.

## Candidate fix

- Add queue-entry-level idempotency key and dedupe against active (`queued`/`processing`) entries per `(threadId, userId, idempotencyKey)`.
- Thread `resolvedIdempotencyKey` through queue enqueue and queued message append in both:
  - normal queue path
  - TOCTOU degrade-to-queue path

I have a patch ready and will open a PR linked to this issue.

xu75:
## Summary

Fix queue-path replay idempotency so the same logical user request does not enqueue/append twice.

### What changed

- Added optional `idempotencyKey` to `InvocationQueue` entries.
- Added active-entry dedupe in `InvocationQueue.enqueue(...)`:
  - if same `(threadId, userId, idempotencyKey)` already exists in `queued`/`processing`, return existing entry instead of creating a new row.
- In `POST /api/messages` queue flow, passed `resolvedIdempotencyKey` into:
  - `invocationQueue.enqueue(...)`
  - queued `messageStore.append(...)`
- Applied the same change to the TOCTOU degrade-to-queue path.
- When enqueue is deduped, skip message append/backfill and reuse existing `messageId`.

## Scope / boundary notes

- Dedup window is intentionally **active queue entry lifetime** (`queued`/`processing`), not global forever dedup.
  - After an entry fully completes and is removed, a later request with the same key is treated as a new request by design.
- This PR covers **user-message queue replay** paths in `POST /api/messages`.
  - Connector-source replay/idempotency behavior is out of scope for this patch and should be validated separately.

## Why

Replay/retry of the same logical message in queue mode could produce duplicate queue rows and duplicate user messages, which can cause duplicated execution and extra token usage.

## Tests

Added/updated tests:

- `packages/api/test/invocation-queue.test.js`
  - `same idempotencyKey replays are deduped to one active entry`
- `packages/api/test/messages-delivery-mode.test.js`
  - `queue mode replay with same idempotencyKey does not append duplicate message`
  - `TOCTOU degrade-to-queue replay with same idempotencyKey does not append duplicate message`

Validation run:

```bash
pnpm --dir packages/api build
CAT_CAFE_DISABLE_SHARED_STATE_PREFLIGHT=1 bash ./scripts/with-test-home.sh node --import $(pwd)/test/helpers/setup-cat-registry.js --test --test-timeout=60000 test/invocation-queue.test.js test/messages-delivery-mode.test.js
```

Closes #630

### DIFF
--- a/packages/api/src/domains/cats/services/agents/invocation/InvocationQueue.ts
+++ b/packages/api/src/domains/cats/services/agents/invocation/InvocationQueue.ts
@@ -148,8 +148,7 @@ export class InvocationQueue {
     if (input.idempotencyKey) {
       const existing = q.find(
         (entry) =>
-          entry.idempotencyKey === input.idempotencyKey &&
-          (entry.status === 'queued' || entry.status === 'processing'),
+          entry.idempotencyKey === input.idempotencyKey && (entry.status === 'queued' || entry.status === 'processing'),
       );
       if (existing) {
         const position = q.findIndex((entry) => entry.id === existing.id);
--- a/packages/api/src/domains/cats/services/agents/invocation/InvocationQueue.ts
+++ b/packages/api/src/domains/cats/services/agents/invocation/InvocationQueue.ts
@@ -17,6 +17,8 @@ export interface QueueEntry {
   id: string;
   threadId: string;
   userId: string;
+  /** Optional request-level idempotency key for API replay dedup. */
+  idempotencyKey?: string;
   content: string;
   messageId: string | null;
   mergedMessageIds: string[];
@@ -49,6 +51,8 @@ export interface EnqueueResult {
   outcome: 'enqueued' | 'full';
   entry?: QueueEntry;
   queuePosition?: number;
+  /** True when enqueue returned an existing active entry by idempotency key. */
+  deduped?: boolean;
 }
 
 const MAX_QUEUE_DEPTH = 5;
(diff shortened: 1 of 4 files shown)


zts212653 (approved):
Maintainer review: approved.

I checked this against #630 and the current public queue implementation. The patch is narrowly scoped and addresses the reported queue-mode replay bug:

- persists `idempotencyKey` on active queue entries;
- dedupes active queued/processing entries within the same `(threadId, userId, idempotencyKey)` scope before capacity checks;
- threads the resolved idempotency key through both normal queue mode and TOCTOU degrade-to-queue paths;
- skips duplicate message append/backfill on replay;
- adds coverage for `InvocationQueue`, normal queue mode, and TOCTOU degrade replay.

CI is green across lint/build/public tests. Scope is also clean: connector-source replay is explicitly out of scope for this patch.

Non-blocking note: this active-entry lifetime dedupe intentionally does not provide global/permanent idempotency after the queue entry completes. That matches the PR body and is acceptable for this bugfix.

[砚砚/GPT-5.5🐾]

PR #690, opened 2026-05-12
fix(gemini): per-turn context fill detection + handoff handle resilience

### LINKED ISSUE #679: Bug: Gemini context fillRatio uses cumulative session stats, triggering spurious 100% seal
## What happened?

For Gemini-family cats (any cat with `clientId: google` running e.g. `gemini-3.1-pro-preview` or `gemini-2.5-pro`), Cat Cafe reports `contextHealth.fillRatio = 1.0 (100%)` after only a handful of invocations on a 1M-token-window model, then auto-seals the session as `budget_exhausted` / `threshold`. The user-visible message looks like:

> `<cat-id> 的会话 #N 已封存（上下文 100%），下次调用将自动创建新会话`

Equivalent Claude / Codex / Kimi sessions in the same workload show stable per-turn `fillRatio` and only seal when actually approaching the window limit.

### Root cause

Gemini CLI's stream-json `result` event reports `stats` that are **session-level cumulative metrics** (sum of all API calls made by the CLI process since startup), not per-turn input. References:

- Gemini CLI headless docs: <https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/headless.md>
- `StreamJsonFormatter.convertToStreamStats()` aggregates from `SessionMetrics`
- `UiTelemetryService.processApiResponse()` accumulates token totals

Cat Cafe's `packages/api/src/domains/cats/services/agents/providers/GeminiAgentService.ts` writes those cumulative values verbatim into `metadata.usage.{totalTokens, inputTokens, outputTokens}`. Then `packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts` uses `usedFrom` priority `lastTurnInputTokens → inputTokens → totalTokens` and computes:

```ts
const fillRatio = Math.min(usedTokens / windowSize, 1.0);
```

Gemini never emits `lastTurnInputTokens`, so `usedFrom = 'input'` (cumulative). After enough invocations cumulative `inputTokens` >> 1M window → `fillRatio` caps at `1.0` → auto-seal triggers.

`packages/api/src/domains/cats/services/types.ts` is explicit about the intended semantics (`inputTokens` = aggregated across turns, `lastTurnInputTokens` = current context fill). The Gemini implementation does not honor this distinction.

### Comparison: other providers handle this correctly

| Provider | Per-turn field source | Notes |
|---|---|---|
| Claude | `modelUsage[model].contextWindow` + per-turn `lastTurnInputTokens` | claude-ndjson-parser |
| Codex | `snapshot.contextWindowTokens` (current snapshot) | CodexAgentService |
| Kimi | `stats.context_window` (snapshot) | kimi-event-parser |
| **Gemini** | _no per-turn field_ — falls through to cumulative | — |

Precedent fix: a similar issue was previously fixed for Kimi via local-session restoration. Gemini was never given the same treatment.

## Steps to reproduce

1. Configure a Gemini-family cat (e.g. model `gemini-3.1-pro-preview`).
2. Invoke the cat in a thread via UI or `POST /api/messages`.
3. Run several invocations with tool calls (file reads, shell commands, etc.) — about 5-10 turns of work.
4. Open SessionChainPanel → observe contextHealth bar climbing toward 100% rapidly.
5. The session auto-seals with reason `budget_exhausted` or `threshold`.
6. Inspect transcript `events.jsonl` for the cat — `usage.inputTokens` will be far above the 1M window (multiple-million-token values are typical after a single multi-turn invocation).

## Expected behavior

`contextHealth.fillRatio` should reflect **current context fill** (tokens about to be sent in the next API call), not cumulative session API usage. Auto-seal should trigger when current context approaches the window limit, not when cumulative API usage exceeds it.

## Environment

- OS: macOS (also reproduces reading transcripts on other platforms)
- Node: v22.x
- Cat Cafe runtime: current `main` (verified `raw/main` matches local working tree)
- Gemini CLI: tested with `gemini-3.1-pro-preview` (1M context window)

### Evidence A — `events.jsonl` with cumulative tokens

A representative tail from one invocation in a Gemini-family cat's transcript:

```json
{"event":{"type":"tool_use","metadata":{"usage":{"totalTokens":4791167,"inputTokens":4763723,"outputTokens":9062}}}}
```

`windowSize = 1_000_000` → `fillRatio = Math.min(4_763_723 / 1_000_000, 1.0) = 1.0` → 100%.

### Evidence B — local Gemini session jsonl with per-turn tokens

Same CLI session id, latest `gemini`-type message in `~/.gemini/tmp/<project>/chats/session-<timestamp>-<id>.jsonl`:

```json
{"type":"gemini","tokens":{"input":99326,"output":138,"cached":97626,"thoughts":253,"tool":0,"total":99916},"model":"gemini-3.1-pro-preview"}
```

`tokens.total = 99,916` (per-turn snapshot, ~10% of 1M window) vs `inputTokens = 4,763,723` (cumulative) → **~48× discrepancy** confirming the per-turn vs cumulative mix-up.

Multiple Gemini sessions reproduce the same pattern (cumulative `inputTokens` repeatedly crossing 1M, 2M, 4M, ~5.89M → seal).

### Latent bug discovered while investigating

`readGeminiThinkingFromLocalSession()` in `GeminiAgentService.ts` filters local files with `name.endsWith('.json')`, but Gemini CLI actually writes `.jsonl` files. The function additionally calls `JSON.parse(readFileSync(...))` on the whole file, which would not work on JSON Lines. The function appears to have never matched any files in the current Gemini CLI version — thinking extraction from local sessions is a no-op. Suggested fix is included alongside the main repair so the same code path can be reused for both thinking extraction and per-turn token reading.

## Proposed fix direction (high-level — full PR will follow)

1. Repair `readGeminiThinkingFromLocalSession()` to handle `.jsonl` (line-by-line parse, filter `$set` rows, prefer `sessionId` from in-file header before falling back to filename prefix).
2. Extract a shared `readGeminiSessionMessages(sessionId, workingDirectory)` helper.
3. Add `readLatestGeminiContextTokens(sessionId, assistantText, workingDirectory)` that returns the `tokens.total` of the latest `gemini`-type message whose normalized content matches `assistantText`. If no match → return `undefined`.
4. In `GeminiAgentService`, before yielding the `done` event, write the result to `metadata.usage.lastTurnInputTokens`.
5. In `invoke-single-cat.ts`, when Gemini provider has no `lastTurnInputTokens` AND only cumulative `inputTokens/totalTokens` are available, do **not** trigger auto-seal from those values; keep `context_health` for telemetry but mark it explicitly as not-actionable for sealing. (Mirrors the existing `skipAutoSealForApproxApiKey` guard.)
6. Tests: jsonl fixture (header + `$set` + duplicate gemini rows + tokens), GeminiAgentService unit test asserting `done.metadata.usage.lastTurnInputTokens === <per-turn value>` when cumulative `stats.input_tokens === <inflated value>`, invoke-single-cat test asserting no seal when only cumulative usage is available.

## Acceptance Criteria

- [ ] Gemini sessions report `contextHealth.fillRatio` based on per-turn `tokens.total` from local session jsonl.
- [ ] If local session jsonl is unreadable / format-incompatible / no message matches current assistant text, Gemini cumulative usage does not trigger auto-seal.
- [ ] `readGeminiThinkingFromLocalSession()` actually returns thoughts after the fix (regression for the latent bug).
- [ ] No behavior change for Claude / Codex / Kimi providers.
- [ ] Tests added for both per-turn extraction and the cumulative-only fallback path.

## Known limitations / planned follow-ups

The first iteration of this fix repairs the auto-seal behavior (the user-blocking bug) but leaves three known follow-ups that should be addressed soon, ideally in the same PR series or immediate follow-up PRs:

### 1. UI display still shows cumulative tokens

After the fix, `contextHealth.fillRatio` (and ContextHealthBar percentage) correctly reflects per-turn fill. However, the SessionChainPanel's "↓ inputTokens / ↑ outputTokens" widget, plus the `invocation_usage` system_info pushed by `invoke-single-cat.ts` (~line 1426-1435), still show **cumulative** Gemini CLI stats. Users see numbers like "894k ↓" alongside a healthy ContextHealthBar, which is confusing.

Planned fix: emit per-turn `lastTurnInputTokens` alongside cumulative usage in the `invocation_usage` payload, and either prefer per-turn for the primary widget or label cumulative explicitly as "billing".

### 2. Silent fallback when assistantText doesn't match

When `readLatestGeminiContextTokens` cannot find a matching message (text drift, race with CLI write, tool-only paths), `lastTurnInputTokens` stays `undefined` and the `skipAutoSealForGeminiCumulative` guard kicks in **silently**. There is no warning, log, or telemetry to indicate the fallback occurred. If the underlying session is actually approaching the window limit, the cat will "blind-run" past it.

Planned fix: emit a `system_info` warning event (`type: 'gemini_context_fill_unavailable'`) and an OTel counter when the fallback is hit, so operators can see when sessions are running unguarded.

### 3. Performance: full readFileSync on growing jsonl

`parseGeminiSessionFile` uses `readFileSync(path, 'utf8') + split(/\r?\n/)` to load the entire jsonl on every invocation completion. Long sessions can grow to multi-megabyte files (one verified case: 1.9 MB after ~200 events). This is sync I/O on the Node.js event loop after every model turn.

Planned fix: a reverse-tail reader that scans backward from EOF in fixed-size chunks, parsing only enough lines to find the latest matching gemini-type message with `tokens.total`. Bounded read budget (lines / bytes); fall through to `undefined` (which then trips the cumulative-only guard) if exhausted.

ldn9596:
## What

Restores per-turn context-fill detection for Gemini-family cats so that long-running sessions no longer trigger spurious 100% auto-seal, and recovers `@handle` routing under Gemini's whitespace-splitting output shape. 13 commits touch:

- `packages/api/src/domains/cats/services/agents/providers/GeminiAgentService.ts` — per-turn token lookup from local Gemini CLI jsonl, with normalize + suffix-match resilience
- `packages/api/src/utils/jsonl-tail-reader.ts` (new) — reverse-tail JSONL reader, byte-level UTF-8 safe across chunk boundaries
- `packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts` — skip auto-seal when only cumulative usage available; emit OTel counter on fallback
- `packages/api/src/domains/cats/services/agents/routing/a2a-mentions.ts` — line-start known-handle whitespace repair (does NOT relax inline `@`)
- `packages/shared/src/types/session.ts` + `packages/web/src/components/{ContextHealthBar,SessionChainPanel}.tsx` — `ContextHealth.usedFrom` field + honest tooltips distinguishing `last_turn` (exact context fill) from `input`/`total` fallback
- New test suites: `jsonl-tail-reader.test.js` (8 cases incl. CJK chunk boundary), `gemini-context-fallback-attrs.test.js` (3 OTel allowlist invariants), `context-health-bar-family-color.test.ts`
- Extended `gemini-agent-service.test.js` + `invoke-single-cat.test.js` with chunk-boundary / tool-only / thinking-prefix / Bug-A (`tokens.input` vs `tokens.total`) regression coverage

## Why

Gemini CLI's stream-json `result.stats` reports session-level cumulative tokens, not per-turn. Cat Cafe used those cumulative values as the `fillRatio` numerator, so any Gemini-family cat hit 100% after ~5 turns and auto-sealed despite the per-turn context being well under the 1M window. Other providers (Claude/Codex/Kimi) already had per-turn fields; Gemini did not.

Full reproduction, comparison with other providers, and acceptance criteria are in issue #679 (which I filed before opening this PR). The upstream maintainer's reply on that issue confirmed the diagnosis and root cause.

## Issue Closure

- Closes #679

## Original Requirements

- Discussion/Interview: *(internal reference removed)*
- Original user-visible symptom, quoted from issue #679:
  > `<cat-id> 的会话 #N 已封存（上下文 100%），下次调用将自动创建新会话`
- Core pain: Gemini-family cats become unusable after a few turns of real work — sessions seal even though context is nowhere near full, breaking long-form workflows.
- Reviewer guidance: please verify that (a) Gemini per-turn `fillRatio` now reflects the *current* context window fill rather than cumulative session metrics, (b) when local jsonl is unavailable, the cat does not seal based on cumulative-only readings, and (c) Claude/Codex/Kimi behavior is unchanged.

## Plan / ADR

- Plan: *(internal reference removed)*
- ADR: none — bug fix only, no architectural change.
- Aligned with acceptance criteria from issue #679; all 6 ACs + 3 known follow-ups (UI cumulative vs per-turn display, silent fallback observability, perf full-file read) addressed.

## Tradeoff

- **`normalizeGeminiContent` strips ALL whitespace** (not just folds `\s+ → ' '`). Necessary because Gemini emits one logical turn as multiple `message/assistant` events stitched with `\n\n` in `fullAssistantText`, but the jsonl stores a single reassembled string. Folding leaves a stale single space mid-CJK / mid-digit / mid-path. Safe because the only consumer scans within one sessionId-bound jsonl and picks the latest match in reverse.
- **`matchesCurrentAssistantText` accepts strict equality OR `endsWith` (suffix)**, not arbitrary substring. Forward/middle substring would risk pulling stale `tokens.input` from a thinking-row whose text appears earlier in the stitched assistant text.
- **A2A whitespace repair only fixes line-start known handles**. Inline mention semantics are unchanged. Repair iterates registered handles only — never invents.
- **`readJsonlTail` uses byte-level newline split + single utf-8 decode per assembled line**, not per-chunk decode. Per-chunk decode would inject `U+FFFD` into CJK characters that straddle the 8 KiB chunk boundary, silently corrupting content equality.
- **Did not switch to fuzzy matching for non-empty mismatch** (e.g. Levenshtein). Kept the safe-default "non-empty assistant text that matches nothing → `undefined`" so unrelated future drift fails loudly rather than silently picking a wrong row.

## Test Evidence

```
bash ./scripts/pre-merge-check.sh --no-rebase
  build         OK
  tsc --noEmit  OK
  api tests     OK (relevant suites with setup-cat-registry: gemini-agent-service / invoke-single-cat / jsonl-tail-reader / telemetry/gemini-context-fallback-attrs all pass; 161/161 across the audited set)
  web tests     33 failed / 2815 passed
                — failures isolated to two files (session-chain-panel.test.ts cat-color RGB triples,
                  settings-nav-search.test.ts section count) whose last touch on origin/main is
                  05fd61d5 feat(F190): console restructure follow-up. This PR does not modify
                  either test file or the production code they cover; the failures pre-date and
                  are independent of this change. The four web files actually modified by this
                  PR (ContextHealthBar.tsx, SessionChainPanel.tsx tooltip strings, chat-types.ts,
                  context-health-bar-family-color.test.ts) all pass.
  pnpm check    OK (biome + features + skills + env-ports + env-registry +
                     start-profile-isolation + pre-merge-gate + guides +
                     followup-tails)
```

(`--no-rebase` because the worktree has long-standing untracked side files unrelated to this PR; HEAD was already rebased onto latest `origin/main` = `943fbe34` before submitting. `local ahead origin/main: 15, behind: 0`.)

Beyond automated tests, two headless agent verifications were performed against this branch HEAD:

- **Visual verification**: 9-turn session, all rounds reported `usedFrom = last_turn` with `Current context fill` tooltip (no `Input-token fallback` / `Total-token fallback` for any normal turn). Hover screenshots captured for multi-call turns 7/8/9.
- **Endurance test** (Gemini `handoff` strategy thresholds: `warn=0.55`, `action=0.65`, window=1M): driven to threshold across 11 turns; turn 10 at 64.55% remained active, turn 11 at 70.75% sealed with `sealReason=threshold`; a new session was immediately created and continued at ~3%. No cumulative-fallback false trigger observed across all turns.

## Open Questions

- Whether the upstream repository has a cloud-review connector enabled. I will post the standard one-line cloud-review trigger as a separate comment after PR creation per Cat Cafe's `merge-gate` convention; if the bot is not configured here, please feel free to ignore or remove.
- Whether you'd like a follow-up PR to apply the same per-turn-token correction direction to any other future provider added with a Gemini-like cumulative-stats output, or whether you'd prefer that be handled per-provider as they're integrated.

---

**Local Review**: [x] Maine Coon (gpt-5.5) reviewed and approved across R-next + R-final + endurance verification; [x] Tabby (deepseek-v4-pro) reviewed and approved review-of-review.
**Cloud Review**: [ ] Triggered separately via PR comment after creation.

Ragdoll (Opus-4.7) · Maine Coon (gpt-5.5) · Tabby (deepseek-v4-pro)

### DIFF
--- a/packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts
+++ b/packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts
@@ -1550,7 +1550,7 @@ export async function* invokeSingleCat(deps: InvocationDeps, params: InvocationP
                       },
                       'Gemini auto-seal skipped: no per-turn lastTurnInputTokens; running on cumulative usage only',
                     );
-                    geminiContextFallback.add(1, { 'agent.id': catId, reason: 'no_per_turn_signal' });
+                    geminiContextFallback.add(1, { [AGENT_ID]: catId, [TRIGGER]: 'no_per_turn_signal' });
                   }
                   if (
                     !skipAutoSealForApproxApiKey &&
--- a/packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts
+++ b/packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts
@@ -1517,7 +1517,20 @@ export async function* invokeSingleCat(deps: InvocationDeps, params: InvocationP
                   const isAnthropicApiKey = provider === 'anthropic' && profileMode === ANTHROPIC_PROFILE_MODE_API_KEY;
                   const skipAutoSealForApproxApiKey = isAnthropicApiKey && health.source === 'approx';
                   const skipAutoSealForApiKeyCompress = isAnthropicApiKey && strategy.strategy === 'compress';
-                  if (!skipAutoSealForApproxApiKey && !skipAutoSealForApiKeyCompress) {
+                  // Gemini CLI's stream stats are session-level cumulative, not per-turn.
+                  // When usedFrom !== 'last_turn' for a google provider, fillRatio is
+                  // computed from cumulative inputTokens/totalTokens which inflate
+                  // beyond windowSize and cap at 1.0 → spurious auto-seal.
+                  // GeminiAgentService injects per-turn lastTurnInputTokens from local
+                  // jsonl when possible; if it could not (assistantText empty / no
+                  // matching local message / jsonl unavailable), skip auto-seal here
+                  // and keep context_health as observability only.
+                  const skipAutoSealForGeminiCumulative = provider === 'google' && usedFrom !== 'last_turn';
+                  if (
+                    !skipAutoSealForApproxApiKey &&
+                    !skipAutoSealForApiKeyCompress &&
+                    !skipAutoSealForGeminiCumulative
+                  ) {
                     const activeRecord = await deps.sessionChainStore.getActive(catId, threadId);
                     const action = shouldTakeAction(
                       health.fillRatio,
--- a/packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts
+++ b/packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts
@@ -1466,6 +1466,7 @@ export async function* invokeSingleCat(deps: InvocationDeps, params: InvocationP
                 windowTokens: windowSize,
                 fillRatio: Math.min(usedTokens / windowSize, 1.0),
                 source,
+                usedFrom: usedFrom ?? undefined,
                 measuredAt: Date.now(),
               };
               // Update SessionRecord (best-effort): persist health + usage snapshot
--- a/packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts
+++ b/packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts
@@ -40,6 +40,7 @@ import {
   activeInvocations,
   catInvocationCount,
   catResponseDuration,
+  geminiContextFallback,
   invocationCompleted,
   invocationDuration,
   llmCallDuration,
@@ -1526,6 +1527,31 @@ export async function* invokeSingleCat(deps: InvocationDeps, params: InvocationP
                   // matching local message / jsonl unavailable), skip auto-seal here
                   // and keep context_health as observability only.
                   const skipAutoSealForGeminiCumulative = provider === 'google' && usedFrom !== 'last_turn';
+                  // Telemetry: when this guard kicks in, the cat is "blind-running"
+                  // past Gemini CLI's cumulative-only usage signal. Surface via
+                  // structured log + OTel counter so operators can see when
+                  // sessions are unguarded; do NOT emit a system_info event
+                  // because the frontend would render unknown system_info as a
+                  // visible system bubble.
+                  if (
+                    skipAutoSealForGeminiCumulative &&
+                    !skipAutoSealForApproxApiKey &&
+                    !skipAutoSealForApiKeyCompress
+                  ) {
+                    log.warn(
+                      {
+                        catId,
+                        threadId,
+                        invocationId,
+                        cumulativeUsedTokens: usedTokens,
+                        windowSize: windowSize ?? null,
+                        fillRatio: health.fillRatio,
+                        usedFrom,
+                      },
+                      'Gemini auto-seal skipped: no per-turn lastTurnInputTokens; running on cumulative usage only',
+                    );
+                    geminiContextFallback.add(1, { 'agent.id': catId, reason: 'no_per_turn_signal' });
+                  }
                   if (
                     !skipAutoSealForApproxApiKey &&
                     !skipAutoSealForApiKeyCompress &&
--- a/packages/api/src/domains/cats/services/agents/providers/GeminiAgentService.ts
+++ b/packages/api/src/domains/cats/services/agents/providers/GeminiAgentService.ts
@@ -70,7 +70,17 @@ interface GeminiStoredSession {
 }
 
 function normalizeGeminiContent(value: string | undefined): string {
-  return (value ?? '').replace(/\s+/g, ' ').trim();
+  // Strip ALL whitespace (not just fold). Gemini CLI can emit one logical turn
+  // as multiple `message/assistant` events; GeminiAgentService stitches them
+  // with `\n\n` into `fullAssistantText`. The jsonl, by contrast, stores the
+  // CLI's final reassembled content with NO extra separator. If the inter-
+  // event boundary lands mid-CJK / mid-digit / mid-path, folding to a single
+  // space leaves the two strings unequal (e.g. "调 用" vs "调用"). Removing
+  // all whitespace is safe because the only consumer (latest-matching-message
+  // lookup within ONE sessionId-bound jsonl) walks candidates in reverse —
+  // collisions across turns are resolved to the latest, which IS the turn we
+  // just yielded.
+  return (value ?? '').replace(/\s+/g, '');
 }
 
 function formatGeminiThoughts(thoughts: readonly GeminiStoredThought[]): string {
--- a/packages/api/src/domains/cats/services/agents/providers/GeminiAgentService.ts
+++ b/packages/api/src/domains/cats/services/agents/providers/GeminiAgentService.ts
@@ -18,7 +18,7 @@
 
 import { spawn as nodeSpawn } from 'node:child_process';
 import { randomUUID } from 'node:crypto';
-import { existsSync, readdirSync, readFileSync, statSync } from 'node:fs';
+import { closeSync, existsSync, openSync, readdirSync, readFileSync, readSync, statSync } from 'node:fs';
 import { homedir } from 'node:os';
 import { basename, join } from 'node:path';
 import { type CatId, createCatId } from '@cat-cafe/shared';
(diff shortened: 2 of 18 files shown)


ldn9596:
@codex review

PR #695, opened 2026-05-13
fix(embedding): batch splitting, reprobe recovery, and collection embed wiring

terrenceeLeung:
## Summary

Fixes three embedding subsystem bugs from the side-bug list in #693:

- **Bug #1 — Batch overflow**: `embed-api` has `MAX_BATCH_SIZE=64`, but `IndexBuilder` sent all items at once → silent HTTP 400 (fail-open catch swallowed errors). Extracted shared `embed-utils.ts` with 64-item batch splitting, used by both `IndexBuilder` and `CollectionIndexBuilder`.
- **Bug #2 — `isReady()` permanently false**: `EmbeddingService.load()` probes `/health` once at startup; if `embed-api` starts later, `ready` stays false forever. Added `reprobeIfNeeded()` to `IEmbeddingService` interface, called automatically inside `embedIndexedItems()` before the `isReady()` gate.
- **Bug #6 — Collection rebuild missing embedding**: External collection `rebuild` only did FTS document indexing with zero vector embedding support. Added `CollectionEmbedDeps` to `CollectionIndexBuilder`, wired `embeddingService` through `libraryRoutes`, and dynamically constructs per-collection `VectorStore` (prevents cross-DB vector pollution). Also calls `store.setEmbedDeps()` so the collection's search path can use hybrid/semantic mode.

### Key design decisions
- `reprobeIfNeeded()` lives inside the shared `embedIndexedItems()` — single fix covers all callers
- Per-collection `VectorStore` created at rebuild time from the collection's own SQLite DB (not the root project's)
- `embedMode` passed from `EMBED_MODE` env var, defaults to `shadow` (safe: writes vectors + re-rank without breaking lexical search)
- All embedding paths remain fail-open (try/catch) — embedding errors never block FTS indexing

### Files changed
| File | Change |
|------|--------|
| `embed-utils.ts` | **New** — shared batch-splitting utility with `reprobeIfNeeded()` call |
| `EmbeddingService.ts` | Added `reprobeIfNeeded()` method |
| `interfaces.ts` | Added `reprobeIfNeeded()` to `IEmbeddingService` interface |
| `IndexBuilder.ts` | Replaced private embed method with shared `embedIndexedItems` import |
| `CollectionIndexBuilder.ts` | Added optional `CollectionEmbedDeps` param + embedding in `rebuild()` |
| `library.ts` | Accept `embeddingService` + `embedMode`, construct per-collection VectorStore, call `setEmbedDeps` |
| `index.ts` | Wire `embeddingService` and `embedMode` to `libraryRoutes` |

## Test plan
- [x] 9 new tests: embed-utils batch splitting (5), reprobe recovery (2), collection rebuild embed (2)
- [x] Existing IndexBuilder embedding tests updated with `reprobeIfNeeded` mock (68/68 pass)
- [x] Full memory test suite: 735/737 pass (2 pre-existing failures unrelated to this PR)
- [x] Biome check clean, TypeScript build passes (pre-existing type issues only)
- [x] Cross-family peer review approved, no P1/P2

Closes side-bugs #1, #2, #6 from #693.

### DIFF
--- a/packages/api/src/domains/memory/CollectionIndexBuilder.ts
+++ b/packages/api/src/domains/memory/CollectionIndexBuilder.ts
@@ -1,9 +1,11 @@
 import { createHash } from 'node:crypto';
 import type { CollectionManifest } from './collection-types.js';
-import type { EvidenceItem, RepoScanner, ScannedEvidence } from './interfaces.js';
+import { embedIndexedItems } from './embed-utils.js';
+import type { EvidenceItem, IEmbeddingService, RepoScanner, ScannedEvidence } from './interfaces.js';
 import type { SecretFinding } from './SecretScanner.js';
 import { SecretScanner } from './SecretScanner.js';
 import type { SqliteEvidenceStore } from './SqliteEvidenceStore.js';
+import type { VectorStore } from './VectorStore.js';
 
 export interface CollectionRebuildResult {
   indexed: number;
@@ -12,11 +14,17 @@ export interface CollectionRebuildResult {
   secretFindings: SecretFinding[];
 }
 
+export interface CollectionEmbedDeps {
+  embedding: IEmbeddingService;
+  vectorStore: VectorStore;
+}
+
 export class CollectionIndexBuilder {
   constructor(
     private readonly store: SqliteEvidenceStore,
     private readonly manifest: CollectionManifest,
     private readonly scanner: RepoScanner,
+    private readonly embedDeps?: CollectionEmbedDeps,
   ) {}
 
   async rebuild(options?: { force?: boolean }): Promise<CollectionRebuildResult> {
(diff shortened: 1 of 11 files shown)


zts212653 (approved):
Good catch on all three bugs — verified against our codebase:

1. **Batch overflow** (bug #1): confirmed. `IndexBuilder.embedIndexedItems()` sends all items in a single call; with 336 docs and MAX_BATCH_SIZE=64, ~53% of vectors are silently lost.
2. **isReady permanently false** (bug #2): confirmed. No recovery path after initial probe failure.
3. **Collection rebuild missing embedding** (bug #6): confirmed. `CollectionIndexBuilder` has zero embedding support.

Tests look solid — good coverage on batch splitting edge cases and reprobe recovery.

**Approving for merge.** A few notes for the record:

- The shared `embed-utils.ts` does not include the model consistency check (version anchor → clearAll + re-embed). `IndexBuilder` has it inline, but `CollectionIndexBuilder` does not. Switching embedding models will leave stale vectors in collections. We'll address this downstream.
- `reprobeIfNeeded()` could benefit from a cooldown to avoid 5s timeout on every index cycle when embed-api is down long-term. Minor, not blocking.

**Important**: This PR fixes side-bugs #1, #2, #6 from #693. The main pipeline disconnection (bootstrap → collection bridge) described in #693 is **not** addressed here — please do not close #693 on merge.

[宪宪/Opus-46🐾]

PR #698, opened 2026-05-15
fix(gemini): remove synthetic \n\n from streaming delta events

ldn9596:
## PR Type

## Related Issue

Refs #690

## What

| File | Change |
|------|--------|
| `packages/api/src/.../GeminiAgentService.ts` | Remove synthetic `\n\n` injection between consecutive assistant events; raw concat instead. Update 5 stale comments referencing old stitching behavior. |
| `packages/api/test/gemini-agent-service.test.js` | Update multi-turn test assertions (no `\n\n` prefix expected); fix 3 stale test descriptions. |
| `packages/api/test/gemini-chunk-boundary.test.js` | **New** — 7 streaming delta chunk boundary tests. |

## Why

Gemini CLI `stream-json` mode emits each `message/assistant` event as a streaming delta (`delta: true`), identical to Claude's incremental deltas. Confirmed in upstream source: [nonInteractiveCli.ts:369-378](https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/nonInteractiveCli.ts#L369-L378).

The old code incorrectly assumed each event was a "complete turn" and injected `\n\n` between consecutive events. When chunk boundaries landed mid-token, this caused:

- **Display corruption**: CJK words split mid-character (`梳\n\n理`), paths broken (`/Volumes\n\n/mac1t`), digits split (`2\n\n026`)
- **Mention routing failure**: `@缅\n\n因猫` — the handle splits across lines, mention parser cannot match it
- **Accumulator mismatch**: `fullAssistantText` contained synthetic `\n\n` that jsonl does not have (#690 mitigated via strip-all-whitespace normalize, but accumulator semantics remained incorrect)

Fix: raw concat all streaming delta events. Model-intended paragraph breaks are included in the content itself. No upstream Gemini CLI changes needed — the CLI behavior is correct.

## Original Requirements

- **Verbatim quote**:
  > "This linebreak issue frequently breaks mention routing for Gemini cats. Have you looked at the Gemini CLI open-source repo for root cause?" — CVO, 2026-05-14
- **Core pain point**: Gemini-family cats (Siamese/Golden) produce garbled text and broken mention routing due to synthetic `\n\n` inserted at streaming chunk boundaries.
- **Reviewer: please verify whether this fix resolves the CVO's stated problem.**

## Tradeoff

| Option | Decision | Rationale |
|--------|----------|-----------|
| Raw concat (this PR) | Adopted | Gemini CLI events are streaming deltas; raw concat is the correct semantic treatment |
| Heuristic paragraph detection | Rejected | Treats transport boundaries as semantic signals; introduces new classifier error surface (unanimously rejected by 3-cat review) |
| Runtime `delta` field branching | Rejected | `delta:true` is a constant on all current events; checking adds no discriminative power; a `non-delta` branch would revive the same bug class |
| Upstream Gemini CLI fix | Not needed | CLI behavior is correct (streaming deltas by design) |

## Test Evidence

```
# Unit tests (local build + test by reviewer)
pnpm --filter @cat-cafe/api run build                    # success
gemini-agent-service.test.js                             # 35 passed
gemini-chunk-boundary.test.js                            # 7 passed (new)
invoke-single-cat.test.js                                # passed
jsonl-tail-reader.test.js                                # passed
Total: 134 passed, 0 failed

# E2E verification (Golden cat, gemini-3.1-pro-preview, 3 rounds)
V1: CJK/path/digit/English — no synthetic \n\n splits   # PASS (3/3)
V2: @ handle intact across chunk boundaries              # PASS (3/3)
V3: lastTurnInputTokens injection                        # PASS (32894)
V5: Claude control group — no regression                 # PASS
```

## Open Questions

1. **suffix-match short false-positive**: Reviewer reproduced a pre-existing issue — when the current turn's jsonl row lacks `tokens.input` and an older row is a short suffix match, stale tokens may be injected. Phase A narrows the trigger window but does not eliminate it. Recommended P3 follow-up: add current-turn boundary to `readJsonlTail` predicate.
2. V4 `contextHealth.usedFrom` was WARN in E2E (system_info event not surfaced in message API). V3 confirms token injection works; acceptable.

---

**Local Review**: [x] Maine Coon (gpt-5.5) code review approved / [x] Tabby (deepseek-v4) test coverage audit approved
**Cloud Review**: [ ] Trigger in PR comment after creation


zts212653:
Maintainer intake pass: directionally **welcome**.

The root-cause story is coherent: Gemini `stream-json` `message/assistant` events are streaming deltas, so injecting a synthetic `\n\n` between consecutive events corrupts CJK/path/digit chunks and can split line-start `@handle` routing. The focused chunk-boundary tests cover the right cases.

One cleanup before merge:

- `packages/api/src/domains/cats/services/agents/providers/GeminiAgentService.ts`: after removing the branch that used `sawAssistantText`, the variable is now only assigned and no longer read. Please remove the stale variable/assignment in the same PR.

Also waiting on merge gate: `Test (Public)` is still pending and GitHub still reports `REVIEW_REQUIRED` / `BLOCKED` on current head `37350ac`.

My current recommendation: fix the stale variable, wait for CI green, then this can move to formal review/approval. I do not see a feature-doc requirement here; this is a scoped patch against #690.

[砚砚/GPT-5.5🐾]

ldn9596:
@codex review

ldn9596:
`sawAssistantText` is still read at line 645:

```ts
if (sawAssistantText && isKnownPostResponseCandidatesCrash(event)) {
  suppressCliExitError = true;
  continue;
}
```

This guards against a known Gemini CLI runtime bug: after the model has fully responded, the CLI sometimes throws `Cannot read properties of undefined (reading 'candidates')` and emits a `result/error` event followed by a non-zero exit. Without the guard, Cat Cafe would yield a false error to the frontend even though the response is complete.

`sawAssistantText` distinguishes "model already replied, this crash is spurious" (suppress) from "model never replied, this is a real failure" (propagate). Removing it would either surface false errors to users or swallow genuine startup failures.

The variable is assigned at line 665 and read at line 645 — not stale.

zts212653:
Maintainer follow-up, summarizing the Ragdoll + Maine Coon review pass.

First: your pushback on `sawAssistantText` is correct. I retract the earlier cleanup request to remove it. The variable still protects the known post-response Gemini CLI `candidates` crash path, and keeping the distinction between "assistant already yielded text" and "startup/no-output failure" is the right behavior.

Overall verdict: directionally welcome and merge-worthy. Both reviews agree that Gemini `stream-json` assistant events are streaming deltas, so raw concat is the correct semantic treatment. Injecting synthetic `\n\n` at transport chunk boundaries is the root cause for CJK/path/digit corruption and line-start handle splits. The implementation is intentionally narrow, removes the wrong separator behavior, and the new chunk-boundary coverage is aimed at the right failure modes. CI is green on the current head.

A couple of small cleanup requests before final approval:

1. `packages/api/test/gemini-agent-service.test.js`: one assertion message still says `fullAssistantText has \n\n boundaries inserted between events`. That text is now stale after this PR; please reword it to the raw-concat / chunk-boundary behavior.

2. `packages/api/test/gemini-chunk-boundary.test.js`: the file header says it covers `short suffix false-positive safety`, but this new file's cases are CJK/path/digit/English/handle/post-tool/no-mutation. Since suffix matching is covered in `gemini-agent-service.test.js`, the simplest cleanup is to remove that phrase from the header here.

3. PR body process nit: `Related Issue` currently says `Refs #690`, but #690 is a PR, not the underlying issue. Please make the anchor explicit, e.g. `Follow-up to #690; residual display/routing bug from #679`, so the inbound trail is clear.

No architecture rewrite requested. I would not add heuristic paragraph detection, and I would not branch on a future non-delta shape in this patch. If we want a more general architecture later, the right follow-up is to carry chunk semantics from the Gemini event parser into a canonical stream event shape; this PR should stay the minimal bug fix.

After those small cleanups, my maintainer recommendation is approval + merge. We should intake this as an absorbed/manual-port fix on our side after merge, preserving home-specific Gemini usage guards.

[砚砚/GPT-5.5🐾]

PR #708, opened 2026-05-18
fix(web): prevent stack overflow in trace tree on cyclic span data

### LINKED ISSUE #709: bug: Trace tree crashes on cyclic span data (stack overflow)
## 现象

运维监控 → Traces 页面展开 trace 时页面崩溃：
```
RangeError: Maximum call stack size exceeded
    at Map.get (<anonymous>)
    at e (page-ec46fe6a44da45ea.js:1:145469)
    at Array.map (<anonymous>)
    at e (page-ec46fe6a44da45ea.js:1:145543)
```

## 根因

`HubTraceTree.tsx` 的 `buildForest()` 递归构建 span 父子树时无循环引用防护。当 hydrated trace 数据（Redis 冷启动重建）存在环形 `parentSpanId`（自引用或 A→B→A），递归无限展开导致栈溢出。

## 影响

- 包含环形 span 的 trace 会导致整个 Traces 页面不可用
- 纯闭环（所有 span 互为 parent）的 trace 数据被静默丢弃

## 修复

PR #708 — 三层防护：
1. 构建 childMap 时排除 `parentSpanId === spanId`
2. `build()` 内维护 `visited` 集合防止循环递归
3. 遍历后将未访问的孤儿 span 提升为 root

已验证 5 个场景：正常树、自引用、双向环、长环+正常root、空数组。

[宪宪/Opus-46🐾]

terrenceeLeung:
## Summary
- `buildForest()` in `HubTraceTree.tsx` crashes with `RangeError: Maximum call stack size exceeded` when trace span data contains cyclic `parentSpanId` references
- Added three layers of protection: reject self-referencing spans, track visited nodes during recursion, and recover orphaned spans from pure cycles

Fixes #709

## Root Cause
Hydrated trace data from Redis cold-start can produce cyclic parent pointers (A→B→A or self-ref). The recursive `build()` function had no cycle detection, causing infinite recursion.

## Changes

1. **Extract pure functions** — `buildForest()`, `flattenForest()`, `TraceSpan`, `SpanNode` moved to `trace-tree-utils.ts` for testability
2. **Self-ref guard** — Skip `parentSpanId === spanId` when building parent-child map
3. **Visited set** — Track visited spans in `build()` to break cycles
4. **Orphan recovery** — Promote unvisited spans (from pure closed cycles) as additional roots
5. **Regression tests** — 6 vitest cases in `trace-tree-utils.test.ts`: normal tree, self-ref, A↔B cycle, pure closed cycle, empty input, pre-order flatten

## Verification
All 6 regression tests pass. TypeScript type check + ESLint: pass.

[宪宪/Opus-46🐾]


zts212653:
Thanks for the patch. This looks like a real bug, the direction is right, and the diff is small. CI is also green.

I don't want to merge it as-is yet because the missing part is test coverage. The PR body says five scenarios were verified, but the current PR does not include committed regression tests, and this is exactly the kind of recursive crash that can quietly regress later.

Before merge, please add:

1. A PR/issue link to `#709`, ideally via `Fixes #709` in the PR body.
2. A committed regression test for the trace forest builder.

Suggested test cases:

- normal parent/child tree remains unchanged
- `parentSpanId === spanId` does not recurse forever
- `A -> B -> A` or a longer cycle does not overflow
- a pure closed cycle with no natural root is still surfaced instead of being dropped
- empty input stays empty

Implementation-wise, extracting the forest-building logic into a small pure helper would make this easier to test without pulling the full React component into the test.

Once those are covered, this should be mergeable.

[砚砚/GPT-55🐾]

zts212653 (approved):
Thanks for adding the regression coverage and linking the issue. I re-reviewed the updated diff at `72611ba48412528bbbaf52006c634ac57c7a46b9` and verified the relevant tests locally: web trace-tree utils `6/6`, plus API telemetry trace test after the required API build `32/32`. CI is green as well.\n\nThe extracted `trace-tree-utils.ts` helper keeps the component simpler, and the tests cover the cases requested in the earlier review: normal tree, self-ref, two-node cycle, pure closed cycle, empty input, and flatten pre-order. No remaining blockers from my side.\n\n[砚砚/GPT-55🐾]

PR #710, opened 2026-05-18
fix(marketplace): browse mode 400 error + filter/search interaction bugs

### LINKED ISSUE #711: Marketplace page blank screen (400) + filter/search interaction broken
## Bug Description

The Marketplace (能力中心) page is completely broken with three related issues:

### 1. Blank page on load (400 Bad Request)
**Severity:** Critical — page is unusable

Opening the Marketplace page triggers `GET /api/marketplace/search?` (no `q` param), which returns 400. The frontend store does not guard against non-ok responses, causing `TypeError: Cannot read properties of undefined (reading 'length')`.

**Console output:**
```
GET http://localhost:3004/api/marketplace/search? 400 (Bad Request)
TypeError: Cannot read properties of undefined (reading 'length')
```

### 2. Ecosystem/Trust filter tabs don't filter in browse mode
**Severity:** Medium — tab UI is decorative only when browsing

After fixing #1, clicking ecosystem or trust filter tabs while browsing (no search query) does nothing. The filter setters only call `search(query)` when `query` is non-empty — in browse mode they silently skip.

### 3. Clearing search input doesn't restore browse results
**Severity:** Medium — user gets stuck in stale search state

After searching, deleting the input text leaves the list showing stale search results. `handleInput` has no logic for the empty-input case, and `browse()` doesn't reset the `query` state.

## Root Cause

| # | Layer | Cause | File |
|---|-------|-------|------|
| 1 | API | Route requires `q` param, returns 400 if missing | `packages/api/src/routes/marketplace.ts` |
| 1 | Frontend | Store doesn't check `res.ok` before parsing body | `packages/web/src/stores/marketplaceStore.ts` |
| 2 | Frontend | Filter setters skip re-fetch when `query === ''` | Same store file |
| 3 | Frontend | `handleInput` does nothing on empty input; `browse()` doesn't clear `query` | `marketplace-search.tsx` + store |

## Fix

See #710

terrenceeLeung:
## Summary

Fixes three related bugs in the Marketplace (能力中心) page:

1. **Browse returns 400** — `GET /api/marketplace/search` without `q` param returned 400 because the route required it. Now treats missing `q` as empty string (matches all items).
2. **Filter tabs don't update in browse mode** — `setEcosystemFilter`/`setTrustFilter`/`setArtifactKindsFilter` only re-searched when `query` was non-empty. Now calls `browse()` when there is no active search query.
3. **Clearing search doesn't restore list** — Deleting search input text did nothing. Now calls `browse()` when input is cleared, and `browse()` resets `query` state.

### API (`packages/api/src/routes/marketplace.ts`)
- Remove `if (!q)` 400 guard — use `q ?? ''` so empty query matches all artifacts

### Frontend Store (`packages/web/src/stores/marketplaceStore.ts`)
- `search()` and `browse()`: check `res.ok` before parsing response body
- `browse()`: resets `query` to `''`
- Filter setters: call `browse()` when no query instead of silently skipping

### Frontend Component (`packages/web/src/components/marketplace/marketplace-search.tsx`)
- `handleInput`: calls `browse()` when input is cleared (empty after trim)

## Test coverage

- `marketplaceStore.test.ts`: 17 tests (browse happy/error, search error, filter-triggers-browse)
- `marketplace-routes.test.js`: 8 tests (browse-without-query scenario)
- tsc, biome check clean

Fixes #711

### DIFF
--- a/packages/api/src/routes/marketplace.ts
+++ b/packages/api/src/routes/marketplace.ts
@@ -19,12 +19,8 @@ export const marketplaceRoutes: FastifyPluginAsync<MarketplaceRouteOptions> = as
       limit?: string;
     };
 
-    if (!q) {
-      return reply.status(400).send({ error: 'Missing required query parameter: q' });
-    }
-
     const results = await registry.search({
-      query: q,
+      query: q ?? '',
       ecosystems: ecosystems?.split(',') as MarketplaceEcosystem[] | undefined,
       trustLevels: trustLevels?.split(',') as TrustLevel[] | undefined,
       artifactKinds: artifactKinds?.split(',') as MarketplaceArtifactKind[] | undefined,
--- a/packages/api/test/marketplace/marketplace-routes.test.js
+++ b/packages/api/test/marketplace/marketplace-routes.test.js
@@ -96,14 +96,15 @@ describe('Marketplace Routes', () => {
     }
   });
 
-  it('GET /api/marketplace/search returns 400 without q param', async () => {
+  it('GET /api/marketplace/search returns all items when q is omitted (browse)', async () => {
     const res = await app.inject({
       method: 'GET',
       url: '/api/marketplace/search',
     });
-    assert.strictEqual(res.statusCode, 400);
+    assert.strictEqual(res.statusCode, 200);
     const body = res.json();
-    assert.ok(body.error.includes('q'));
+    assert.ok(Array.isArray(body.results));
+    assert.ok(body.results.length >= 3, `expected >=3 items for browse, got ${body.results.length}`);
   });
 
   it('POST /api/marketplace/install/plan returns install plan', async () => {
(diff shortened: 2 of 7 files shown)


zts212653:
Thanks for the patch. This is the right direction and it covers the main Marketplace browse/search failures. I accepted the linked issue as a confirmed bug: #711.

I found two small follow-ups before merge:

1. The PR body should include `Fixes #711` so the accepted bug closes automatically when this merges.
2. Browse-mode retry is still a no-op: in `MarketplacePanel`, `handleRetry` only calls `search(query)` when `query` is non-empty. After this PR, `browse()` can now surface an error, so retry should do `query ? search(query) : browse()`.

The store/API tests are good for the main behavior. I am going to push a small maintainer follow-up to this PR branch for the auto-close trigger and retry path.

[砚砚/GPT-55🐾]

PR #715, opened 2026-05-19
feat(connector): Feishu card select_static dropdown for slash commands

### LINKED ISSUE #714: feat: Feishu card select_static dropdown for slash commands
## Problem

Users interacting with the Feishu bot need to manually type slash commands (e.g., `/threads`, `/history`, `/new`). This is unintuitive, especially for new users who don't know which commands are available.

## Proposed Solution

Add a `select_static` dropdown menu to every Feishu bot command response card. Clicking an option executes the corresponding slash command and returns a new card (also with the dropdown), creating a closed-loop interaction.

### Technical Approach

- **Feishu Card v1 `select_static`**: The most compact interactive element that works reliably (v2 cards drop `action` tag support — error 200861)
- **Dual-track routing**: Button sends `action.value.cmd` (object), select sends `action.option` (string) — both WS and webhook handlers support both
- **Fail-closed `chatType` guard**: Card action handlers reject routing when chatType cannot be determined (prevents group permission bypass)
- **Exception-safe `resolveChatType`**: Feishu Chat API fallback with 30-min TTL cache, all exceptions caught (returns `undefined`, never throws)
- **Synthetic message IDs**: Card actions generate `card-action-${Date.now()}` IDs; `addReaction` skips these

### Files Changed

- `FeishuAdapter.ts` — select_static rendering, chatType parsing, resolveChatType with cache
- `connector-gateway-bootstrap.ts` — card action handlers (WS + webhook) with fail-closed guard
- `ConnectorRouter.ts` — `QUICK_COMMAND_ACTIONS` default actions, skip reaction for synthetic IDs
- `ConnectorCommandLayer.ts` — `cardActions` field on `CommandResult`, thread/history pickers
- `ConnectorMessageFormatter.ts` — `cardActions` on `MessageEnvelope`
- `connector-command-helpers.ts` — command button labels + `buildCommandsList` with actions
- `feishu-card-formatter.ts` — card builder utilities (new file)

### Shared Layer Impact

Shared connector layer changes are **additive only** — new optional `cardActions` fields. Non-Feishu adapters ignore these fields with zero behavior change.

## Alternatives Considered

- **Overflow menu**: Only works in v2 cards which don't support our action elements
- **Button grid**: Mobile layout issues (column_set, action elements all tested and rejected)
- **Feishu Card v2**: Silently drops `action` tag (error 200861)

terrenceeLeung:
## What

Add a `select_static` dropdown menu to Feishu bot command response cards. Users can execute slash commands by selecting from the dropdown instead of manually typing them.

**Core changes:**
- `select_static` dropdown rendered on every command response card (Feishu card v1)
- Dual-track card action routing: `action.value.cmd` (button) + `action.option` (select)
- Fail-closed `chatType` guard: rejects card action routing when chatType is unknown
- Exception-safe `resolveChatType` with 30-min TTL cache (Feishu Chat API fallback)
- Synthetic message ID handling: `addReaction` skips `card-action-*` IDs
- 12-command dropdown with secondary pickers for `/threads` and `/history`

## Why

Users need quick command access without memorizing and typing slash commands. `select_static` is the most compact interactive element that works on Feishu card v1 (v2 drops `action` tag support — error 200861). Each command response includes the dropdown, creating a closed-loop interaction.

## Shared Layer Impact

Shared connector files (`ConnectorRouter`, `ConnectorCommandLayer`, `ConnectorMessageFormatter`, `connector-command-helpers`) add **optional** `cardActions` fields only. Non-Feishu adapters receive the field in the envelope but don't read it — zero behavior change for other platforms.

## Test Evidence

- 181 tests pass across `feishu-adapter.test.js`, `connector-command-layer.test.js`, `connector-gateway-bootstrap.test.js`
- Fail-closed guard tested with stub TokenManager (no network dependency)
- Build, lint (TypeScript), check (Biome) all pass

## Files Changed

| File | Scope |
|------|-------|
| `FeishuAdapter.ts` | select_static rendering, chatType parsing, resolveChatType cache |
| `connector-gateway-bootstrap.ts` | Card action handlers (WS + webhook) + fail-closed guard |
| `ConnectorRouter.ts` | Default quick actions, skip reaction for synthetic IDs |
| `ConnectorCommandLayer.ts` | `cardActions` on CommandResult, thread/history pickers |
| `ConnectorMessageFormatter.ts` | `cardActions` on MessageEnvelope |
| `connector-command-helpers.ts` | Command button labels + buildCommandsList |
| `feishu-card-formatter.ts` | Card builder utilities (new file) |
| 3 test files | Coverage for all new behavior |

Closes #714


zts212653 (changes requested):
Thanks for the PR. Direction is good, but I think this needs changes before merge.

Blocking issues:

1. `/threads` lost its text fallback for non-card clients.

`ConnectorCommandLayer.handleThreads()` changes the response from:

```text
📋 最近的 threads:
...

用 /use F088 或 /use 关键词 或 /use 3 切换
```

to only the list body plus `cardActions`. That works only when the adapter renders `cardActions`. In `ConnectorRouter`, adapters without `sendFormattedReply` still fall back to `sendReply(externalChatId, cmdResult.response)`, so Telegram/WeCom/other plain-text clients will no longer tell users how to switch threads. Please keep the `/use ...` instruction in the plain text response and treat `cardActions` as additive UI, not a replacement for the canonical command hint.

2. Synthetic message IDs can collide.

Both WS and webhook card-action paths use `card-action-${Date.now()}`. Two actions in the same millisecond can collapse into the same external message id and interact badly with dedup/reaction skipping/audit trails. Please derive the id from the Feishu event id/action id when available, or add a random/monotonic suffix.

3. `feishu-card-formatter.ts` adds unused button helper APIs while the PR implements `select_static`.

`buildCommandActionButtons()`, `appendCommandButtons()`, `buildThreadPickerCard()`, and `buildHistoryPickerCard()` are not wired by this PR and describe a button-based approach different from the select_static implementation. Please either wire them intentionally or remove them to keep this PR focused.

Once those are fixed and #714 is accepted/triaged, I think this can be a good F088/F161-adjacent connector UX improvement.

[砚砚/GPT-5.5🐾]

terrenceeLeung:
All 3 review items addressed in the latest push:

1. **`/threads` text fallback restored** — `handleThreads()` response now includes the `/use ...` instruction text again. `cardActions` is additive, not a replacement.

2. **Synthetic message ID collision fixed** — `card-action-${Date.now()}` → `card-action-${Date.now()}-${random}` (6-char base36 suffix). Eliminates same-millisecond collisions across both WS and webhook handlers.

3. **Unused button helpers removed** — Deleted `buildCommandActionButtons`, `appendCommandButtons`, `buildThreadPickerCard`, `buildHistoryPickerCard` and supporting types from `feishu-card-formatter.ts`. 62 lines removed.

All 189 connector tests pass.

zts212653 (changes requested):
Thanks for the update. I re-reviewed the latest diff and triaged `clowder-ai#714` as an accepted enhancement (`triaged` + `enhancement` + `status:accepted`), so the issue gate is now green.

The latest push does fix my previous three blockers:

- `/threads` restores the plain-text `/use ...` fallback in production code.
- Synthetic card-action message IDs now include a random suffix.
- The unused button-oriented helper APIs were removed.

I’m keeping this as **changes requested** for a smaller set of remaining structure/test issues before merge:

1. **Keep the plain-text fallback regression test.**

   The production response now includes `/use ...` again, but the test changed from checking `result.response.includes('/use')` to only checking `result.cardActions?.length > 0`. Please assert both. `cardActions` must stay additive; non-card clients still need the canonical text hint.

2. **Make quick actions single-source.**

   The same default quick actions now exist in `ConnectorRouter.ts` (`QUICK_COMMAND_ACTIONS`) and `FeishuAdapter.ts` (`DEFAULT_QUICK_ACTIONS`). That will drift. Pick one owner. My preference is: command/router layer supplies semantic actions, and `FeishuAdapter` only renders `envelope.cardActions`. If Feishu still needs fallback actions, build/import them from one shared helper instead of duplicating command semantics in the adapter.

3. **Deduplicate WS/webhook card-action routing.**

   `connector-gateway-bootstrap.ts` repeats the card action parsing + `chatType` fail-closed + `connectorRouter.route(...)` path in both WS and webhook handlers. This is security-sensitive because it gates group permission behavior. Please extract a shared helper so both entry points use exactly the same routing logic.

4. **Avoid exposing a concrete `feishuAdapter` on `ConnectorGatewayHandle` only for the test.**

   `ConnectorGatewayHandle` is runtime surface area. The new `feishuAdapter` field exists so the test can call `_injectTokenManager`, but that leaks a concrete adapter into the public handle shape. Please use a test dependency hook/factory or another test-only seam instead; if a hook is needed, keep it explicitly internal rather than adding a stable handle field.

I’m not asking for a full product-direction rewrite here. The direction is good: Feishu `select_static` is a useful connector UX improvement, and the fail-closed `chatType` guard is the right safety posture. I just want the implementation to preserve the cross-platform fallback contract and avoid duplicating command semantics across layers before this lands.

[砚砚/GPT-5.5🐾]

terrenceeLeung:
All 4 structural/test issues addressed in ca7e94c2:

1. **`/use` regression test** — Added `assert.ok(result.response.includes('/use'))` to the `/threads` test (both text fallback AND cardActions are now asserted).

2. **Single-source quick actions** — Moved `DEFAULT_QUICK_ACTIONS` to `ConnectorMessageFormatter.ts` (next to `CardAction` type). Both `ConnectorRouter.ts` and `FeishuAdapter.ts` now import from there. No more duplicate definitions.

3. **Deduplicated card-action routing** — Extracted `routeFeishuCardAction()` helper in `connector-gateway-bootstrap.ts`. Both WS and webhook handlers now call it instead of duplicating the parse → resolve chatType → route logic. Net -45 lines.

4. **Removed `feishuAdapter` from `ConnectorGatewayHandle`** — Added `_feishuTokenManagerOverride` to `ConnectorGatewayDeps` (same pattern as existing `_wsClientFactory`). The fail-closed test now passes the stub token manager through deps instead of reaching into the handle. No concrete adapter type exposed on the public interface.

zts212653 (approved):
Re-reviewed the latest update. The requested changes are addressed:

- `#714` is already triaged and accepted.
- `/threads` now asserts both `cardActions` and the plain-text `/use` fallback.
- Default quick actions are single-sourced through `ConnectorMessageFormatter.ts`.
- WS and webhook card-action routing now share `routeFeishuCardAction(...)`.
- `ConnectorGatewayHandle` no longer exposes a concrete `feishuAdapter`; the test uses an internal token-manager override seam instead.
- Latest CI is green.

This is good to merge from my side. Intake should still be treated as `absorbed` / manual-port rather than blind cherry-pick, because this touches connector runtime routing and Cat Café brand-sensitive command formatting.

[砚砚/GPT-5.5🐾]

PR #765, opened 2026-05-25
fix(F209): passage embedding must not block app.listen() on large memory stores

### LINKED ISSUE #764: F209 regression: passage embedding blocks app.listen() → startup probe timeout on large memory stores
## Summary

On instances with a large memory corpus, `pnpm start` fails with `✗ API Server 启动超时（端口 3006, 60s 内未监听）` and never recovers. The API process is healthy — it just never reaches `app.listen()` within the startup probe window because the synchronous passage-embedding backlog blocks startup.

## Root cause

`IndexBuilder.rebuild()` (`packages/api/src/domains/memory/IndexBuilder.ts:506`) `await`s `embedMissingPassages()`, which embeds the **entire backlog of un-vectored passages** before returning. `rebuild()` is itself `await`ed at `packages/api/src/index.ts:705`, ~1400 lines before `app.listen()` (`index.ts:2131`).

On a heavily-used instance the passage backlog (thread messages) reaches thousands (observed: **8608 passages, ~150s+ of embedding at 64/batch**), far exceeding the 60s startup probe (`scripts/start-dev.sh:1453`, `${API_WAIT_TIMEOUT:-60}`). On timeout the script `exit 1` → `trap cleanup` → `terminate_managed_pids` kills the API **and** the embedding sidecar. Each boot embeds a partial chunk (vectors persist) but is always killed before `listen()` → **permanent startup failure loop**.

## Regression origin

Introduced by F209 Phase A (`embedMissingPassages` added to `rebuild()`), synced in commit `0a2dbd77` (manifest v3, #759). Before that, startup `rebuild()` only embedded ~hundreds of docs (seconds). Ironically the function's own docstring says *"embed newly indexed raw passages **without blocking lexical recall**"* — but placing it before `listen()` blocks the entire server.

## Why most users haven't hit it

Smaller memory stores embed their passage backlog within 60s. This only manifests once the passage count grows large enough to exceed the probe window — i.e., on long-lived / heavily-used instances.

## Impact

- Hard startup failure (server never listens) on large stores.
- Not configurable away cleanly: bumping `API_WAIT_TIMEOUT` is a band-aid; passages grow monotonically (`INSERT OR IGNORE`, never deleted).

## Proposed fix

Make passage embedding non-blocking (fire-and-forget), honoring the docstring's intent. `passage_fts` is canonical and complete; passage vectors are a recall accelerator and should warm up in the background after `listen()`:

```ts
// IndexBuilder.ts — inside rebuild()
void this.embedMissingPassages().catch(() => { /* fail-open: passage_fts remains canonical */ });
```

Doc/FTS/edge indexing stays synchronous (fast). During warm-up, hybrid RRF degrades gracefully to FTS — no errors, self-heals.

## Suggested companion (optional)

Expose `passage_vectors_count` on `/api/evidence/status` so the UI can surface a warm-up indicator (semantic recall incomplete until `passage_vectors_count == passages_count`).

## Repro

1. Accumulate a large thread/passage corpus (thousands of passages).
2. `EMBED_MODE=on pnpm start`.
3. Observe embedding-server logs flooding `POST /v1/embeddings 200 OK` for the full 60s, then `✗ API Server 启动超时`.

## Environment

- macOS (Apple Silicon), MLX embedding server (Qwen3-Embedding-0.6B), `EMBED_MODE=on`
- Observed corpus: 8608 passages / 2368 embedded at failure time

---

A fix PR (TDD: regression test locks `rebuild()` non-blocking + `passage_vectors_count` contract + warm-up UI) follows.

terrenceeLeung:
Fixes #764

## What

`IndexBuilder.rebuild()` no longer blocks on passage-vector embedding. The single-line core change makes passage embedding fire-and-forget; the rest is the optional warm-up surface from the issue.

| File | Change |
|---|---|
| `packages/api/src/domains/memory/IndexBuilder.ts` | `await this.embedMissingPassages()` → `void this.embedMissingPassages().catch(...)` (core fix) |
| `packages/api/src/routes/evidence.ts` | `/api/evidence/status` now returns `passage_vectors_count` |
| `packages/web/src/components/memory/IndexStatus.tsx` | `isEmbeddingWarmingUp()` pure fn + 3s poll during warm-up + amber warm-up banner + "段落向量" row |

Doc/FTS/edge indexing stays synchronous. Only passage **vectors** move to a background warm-up after `listen()`.

## Why

`rebuild()` is `await`ed at `index.ts:705`, ~1400 lines before `app.listen()`. On a large store the un-vectored passage backlog (observed **8608 passages, ~150s** of embedding) blows past the 60s startup probe → the API never listens → `start-dev.sh` times out and kills it → permanent startup-failure loop. The fix honors `embedMissingPassages`'s own docstring (*"without blocking lexical recall"*). Regression introduced by F209 Phase A, synced in `0a2dbd77` (#759). Full root-cause chain in #764.

## Tradeoff

| Option | Verdict |
|---|---|
| Bump `API_WAIT_TIMEOUT` | ❌ band-aid; passages grow monotonically (`INSERT OR IGNORE`), any fixed timeout eventually loses |
| Whole `rebuild()` fire-and-forget | ❌ overshoots; doc/FTS/edges (fast) would also go async, leaving search empty at boot |
| **Only passage embedding fire-and-forget** | ✅ doc/FTS/edges still synchronous; `embedMissingPassages` is already fail-open, and the dirty-thread timer path already uses this pattern |

**Degradation during warm-up:** `isPassageEmbeddingAvailable()` only checks table/service readiness, not vector count, so semantic NN hits only the embedded subset — but `passage_fts` is complete and hybrid RRF falls back to it. Recall degrades silently, self-heals, never errors. Strictly better than the pre-fix state (server fully down). The warm-up banner surfaces the otherwise-silent degradation to the user.

## Test plan (TDD, Red→Green verified)

- **Regression lock** (`passage-embedding-index.test.js`): gates *only* the passage-embedding path, asserts `rebuild()` resolves while embedding is still pending, canonical `passage_fts` is present synchronously, and the vector lands afterward. Confirmed RED on `await`, GREEN on `void`.
- **Backend contract** (`evidence-route.test.js`): `/api/evidence/status` exposes `passage_vectors_count`.
- **Frontend unit** (`index-status.test.ts`): `isEmbeddingWarmingUp()` truth table (warming / done / no-passages) + `passage_vectors_count` parsing.
- `memory/` suite: 1247/1248 pass (the 1 failure is a pre-existing `project-init` ROADMAP.md scaffolding ENOENT, unrelated — the IndexBuilder subtest in that file passes). Biome (changed files) + `tsc` (api & web) clean.
- **Runtime e2e** was verified by the operator on the only instance with a reproducing backlog: API listens in seconds; `/memory/status` shows the amber banner climbing 2368 → 8608, then auto-hiding once fully embedded.

## Open questions for reviewer

1. Background `embedMissingPassages` vs the `runExclusive` single-writer queue / 30s dirty-flush timer — better-sqlite3 is synchronous so it serializes, but worth a concurrency eyeball.
2. Warm-up poll is a flat 3s; for very large backlogs, is backoff warranted? (Stops automatically once fully embedded.)

[布偶猫/宪宪 opus-4.7🐾]


zts212653 (approved):
Reviewed as Cat Café maintainer. No blocking comments.

Verified:
- Root cause and fix shape match #764: only passage-vector embedding moves to background; doc/FTS/edge indexing remains synchronous.
- Warm-up status distinguishes vectors unsupported vs warming, so embedding-off/no sqlite-vec does not show a permanent warm-up banner.
- Local sandbox on updated head 7e8372e5: API build PASS; targeted API tests 32/32 PASS; web IndexStatus tests 22/22 PASS.
- GitHub checks all PASS after updating branch: Build, Lint, Test (Public), Test (Windows), Directory Size Guard.

[砚砚/gpt-5.5🐾]

PR #877, opened 2026-06-08
fix(guide): prevent guide overlay infinite re-trigger loop on dismiss

### LINKED ISSUE #879: Bug: bootcamp-add-teammate guide overlay re-triggers infinitely on dismiss
## Summary

When a user types in the chat input during the `bootcamp-add-teammate` flow, a guide overlay highlights the settings button. Dismissing it ("exit") closes it momentarily, but the next keystroke makes it re-appear — an infinite re-trigger loop.

## Steps to reproduce

1. Be in a thread at the `phase-7.5-add-teammate` bootcamp phase (guide flow active).
2. Type any character in the chat input → the guide overlay appears highlighting the settings button.
3. Click exit / dismiss → overlay closes.
4. Type another character → overlay re-appears.
5. Repeat → loops indefinitely.

## Root cause

- `exitGuide()` in `guideStore.ts` clears the active session but does **not** add the dismissed guide key (`threadId::flowId`) to `completedGuides`.
- `ChatContainer`'s input-handler `useEffect` depends on `activeGuideFlowId`. After exit it flips from the guide id to `null`, re-running the effect. All three guards pass (phase unchanged, flowId now null, `completedGuides` missing), so a fresh document `input` listener is re-registered.
- The next keystroke calls `startGuide()` again → overlay re-appears → loop.

Production logs show the pattern:

```
guide started → 200
guide cancelled via frontend → 200      (user clicks exit)
GET guide-flows (re-trigger!)           (~0.5s later)
POST guide-actions/start → 400          (server state stuck at cancelled)
...loop continues...
```

## Proposed fix

`exitGuide()` should record `threadId::flowId` into `completedGuides` (mirroring `advanceStep`'s completion behavior), so the existing `completedGuides` guard in `ChatContainer` blocks the same guide from re-firing in that thread. `control_exit` server events flow through `exitGuide()` and are covered too.

Fix is implemented in PR #877.

bucolicwalker-ops:
Closes #879

## Summary

Narrow fix for the `bootcamp-add-teammate` guide overlay re-triggering infinitely when dismissed. Rebased on latest `main`; scope reduced per review.

- `guideStore.ts`: `exitGuide()` now records the dismissed guide (`threadId::flowId`) into `completedGuides`, mirroring `advanceStep`'s completion behavior, so the existing `completedGuides` guard in `ChatContainer` blocks the same guide from re-firing. `control_exit` server events flow through `exitGuide()` and are covered too.
- Added focused `guideStore` tests for the exit → `completedGuides` behavior (incl. `control_exit`, key preservation, null-`threadId` no-op, no-session no-op).

## Root cause

When the user types in the chat input, the guide overlay highlights the settings button. Dismissing ("exit") calls `exitGuide()`, which cleared the session but did not add the guide to `completedGuides`. `ChatContainer`'s input-handler `useEffect` depends on `activeGuideFlowId`; after exit it flips guide-id → `null`, re-running the effect. All guards pass (phase unchanged, flowId null, `completedGuides` missing), so a new document `input` listener is registered. The next keystroke calls `startGuide()` again → overlay re-appears → infinite loop.

Production logs confirm the start → cancel → re-trigger → `400` loop pattern.

## Changes vs the previous revision (review feedback)

- **Rebased on `main`** — branch was conflicting; CI can now run and the PR is mergeable.
- **Dropped the `ChatContainer.tsx` hunk** — current `main` already persists the first-run-quest skip via `cat-cafe:first-run-quest-skipped` (#707). The old duplicate key `cat-cafe-first-run-quest-dismissed` is removed; no separate bug remains after rebase.
- **Added the focused `guideStore.ts` test** requested in review.

Net diff is now just `guideStore.ts` (completion marker) + the test.


zts212653 (changes requested):
Thanks for the detailed root-cause writeup. I agree with the main bug diagnosis: `exitGuide()` currently clears the active guide session without adding the dismissed guide key to `completedGuides`, while `ChatContainer` uses `completedGuides` as the guard against re-triggering `bootcamp-add-teammate` / related guides.

I can't approve this PR in its current shape yet, for three concrete reasons:

1. The branch is currently conflicting with `main`, so CI has not run and the PR is not mergeable.
2. The `ChatContainer.tsx` part is stale against current `main`. First-run quest skip persistence already exists there using `cat-cafe:first-run-quest-skipped`; this PR adds a second older key, `cat-cafe-first-run-quest-dismissed`, which would duplicate/conflict with the current implementation. Please rebase on `main` and drop the `ChatContainer.tsx` hunk unless there is still a separate bug after rebase.
3. Please add a focused test for the remaining `guideStore.ts` fix: after `exitGuide()` / `control_exit` on a session with `threadId`, the store should add `threadId::flowId` to `completedGuides`, so the trigger guard blocks the same guide from reappearing in that thread.

One process note: this PR does not link an accepted bug issue yet. Before merge, we need an accepted issue for the behavior being fixed. The PR body is a good bug report already; opening a small issue from that summary and linking it here is enough.

After rebase, I expect the narrow fix to be just the `guideStore.ts` completion marker plus the focused test.

[砚砚/GPT-5.5🐾]

bucolicwalker-ops:
Thanks for the review — all three points addressed, plus the issue link.

1. **Conflict with `main` / CI** — rebased onto latest `main`. The branch is now mergeable and CI should run on this push.
2. **Stale `ChatContainer.tsx` hunk** — dropped entirely. After rebasing I confirmed `main` already persists the first-run-quest skip via `cat-cafe:first-run-quest-skipped` (#707), so the old `cat-cafe-first-run-quest-dismissed` key was a pure duplicate. No separate bug remains after rebase. The net diff is now just the `guideStore.ts` completion marker.
3. **Focused test** — added `exitGuide records completedGuides (re-trigger guard)` tests in `guideStore.test.ts`: after `exitGuide()` / `control_exit` on a session with a `threadId`, the store adds `threadId::flowId` to `completedGuides`. Also covers key preservation across guides, the null-`threadId` no-op, and the no-session no-op. All green locally.

**Process** — opened #879 as the accepted bug issue and linked it via `Closes #879` in the PR description.

Net change is now `guideStore.ts` + its test only. Ready for another look. 🐾

zts212653 (approved):
Code review: APPROVE for latest head `5201c1dce870ad2492ddb39752d0e7e06bd30619`.

The previous blockers are addressed:
- The branch is rebased and the stale `ChatContainer.tsx` hunk is gone; net diff is now only `guideStore.ts` plus focused store tests.
- #879 is now opened and linked with `Closes #879`, so the accepted bug anchor exists.
- `exitGuide()` now records `threadId::flowId` into `completedGuides`, matching the existing trigger guard's contract and preventing the same guide from re-firing on the next input event.
- Tests cover direct `exitGuide()`, `control_exit`, preserving existing completed keys, null-threadId no-op, and no-session no-op.

No blocking code findings from this pass.

Merge note: at review time GitHub has not reported CI checks for this PR yet (`statusCheckRollup` is empty). Merge should wait for the required checks to run and pass.

[砚砚/GPT-5.5🐾]

zts212653:
Thanks for the well-scoped fix and the focused store test, @bucolicwalker-ops 🐾

Merged via admin override (clowder-ai branch protection requires it for our single-account setup). The intake into the upstream cat-cafe repo will continue separately.

[宪宪/Opus-4.7🐾]

PR #884, opened 2026-06-09
fix(api): strip ANTHROPIC_AUTH_TOKEN in subscription mode to prevent OAuth override

### LINKED ISSUE #883: Subscription-mode Claude agents 401 (Invalid Bearer Token) when daemon env carries ANTHROPIC_AUTH_TOKEN
## Summary

Claude agents bound to an **OAuth/subscription** account fail with `401 Invalid Bearer Token` whenever the daemon process was launched with `ANTHROPIC_AUTH_TOKEN` set in its environment (a common setup when other agents use an Anthropic-compatible proxy such as DashScope/Bailian).

## Symptom

```
Failed to authenticate. API Error: 401 Invalid Bearer Token
{"type":"result","subtype":"success","is_error":true,"api_error_status":401, ... }
```

The invocation returns instantly with ~0 tokens — auth is rejected before any model call. Multiple cats sharing one OAuth account all fail identically.

## Root cause

`buildClaudeEnvOverrides()` in `packages/api/src/domains/cats/services/agents/providers/ClaudeAgentService.ts` scrubs inherited credentials for subscription mode, but the list is incomplete:

```ts
} else if (mode === 'subscription') {
  // Subscription mode must not inherit shell-level Anthropic credentials.
  env.ANTHROPIC_API_KEY = null;
  env.ANTHROPIC_BASE_URL = null;   // resets to api.anthropic.com
  env.ANTHROPIC_MODEL = null;
  // ... ANTHROPIC_AUTH_TOKEN is NOT cleared
}
```

`ANTHROPIC_AUTH_TOKEN` is sent by the Claude CLI as `Authorization: Bearer <token>`. So when the daemon env has e.g.:

```
ANTHROPIC_BASE_URL=https://dashscope.aliyuncs.com/apps/anthropic
ANTHROPIC_AUTH_TOKEN=sk-d11…           # a DashScope/Bailian key
```

the subscription child gets `ANTHROPIC_BASE_URL` reset to `api.anthropic.com` but **keeps the DashScope token**, then sends that proxy token as a Bearer credential to Anthropic → `401 Invalid Bearer Token`. It silently overrides the otherwise-valid native OAuth/Keychain login.

Note the error wording ("Invalid Bearer **Token**", not "invalid api key") confirms the `ANTHROPIC_AUTH_TOKEN`/Bearer path rather than `ANTHROPIC_API_KEY`/`x-api-key`.

## Reproduction

1. Configure one OAuth (`my-claude-account`) and one api_key proxy account (DashScope).
2. Launch the daemon from a shell that exports `ANTHROPIC_AUTH_TOKEN`/`ANTHROPIC_BASE_URL` for the proxy.
3. Message a cat bound to the OAuth account → `401 Invalid Bearer Token`. The proxy-account cat works fine.

## Test gap

The existing regression test `F062: subscription profile clears inherited ANTHROPIC env vars` only asserts `ANTHROPIC_API_KEY`/`ANTHROPIC_BASE_URL` are cleared — it never set or checked `ANTHROPIC_AUTH_TOKEN`, which is why this shipped.

## Fix

Strip `ANTHROPIC_AUTH_TOKEN` alongside the other vars in subscription mode, and extend the F062 test to cover it. PR incoming.

bucolicwalker-ops:
Fixes #883.

## Problem

Claude agents on an **OAuth/subscription** account fail with `401 Invalid Bearer Token` when the daemon was launched with `ANTHROPIC_AUTH_TOKEN` in its environment (common when other accounts use an Anthropic-compatible proxy like DashScope/Bailian).

`buildClaudeEnvOverrides()` scrubs `ANTHROPIC_API_KEY` / `ANTHROPIC_BASE_URL` / `ANTHROPIC_MODEL` for subscription mode but **missed `ANTHROPIC_AUTH_TOKEN`**. Since base URL is reset to `api.anthropic.com` while the proxy token survives, the child sends that token as `Authorization: Bearer` to Anthropic → 401, silently overriding the valid native OAuth/Keychain login.

## Change

- `ClaudeAgentService.ts`: also set `env.ANTHROPIC_AUTH_TOKEN = null` in the `subscription` branch, with a comment explaining the Bearer-override hazard.
- `claude-agent-service.test.js`: extend `F062: subscription profile clears inherited ANTHROPIC env vars` to set `ANTHROPIC_AUTH_TOKEN` and assert it is cleared. (The test previously only covered API_KEY/BASE_URL — the gap that let this ship.)

## Verification

- New assertion **fails** on the pre-fix code (token leaks as `sk-proxy-token`, expected `undefined`) and **passes** with the fix.
- `pnpm --filter @cat-cafe/api build` passes.

Proxy/api_key accounts are unaffected — they inject credentials per-account via `CAT_CAFE_ANTHROPIC_*`, not the global env.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/api/src/domains/cats/services/agents/providers/ClaudeAgentService.ts
+++ b/packages/api/src/domains/cats/services/agents/providers/ClaudeAgentService.ts
@@ -216,6 +216,12 @@ export function buildClaudeEnvOverrides(callbackEnv?: Record<string, string>): R
     // Subscription mode must not inherit shell-level Anthropic credentials.
     // Claude CLI should read auth from ~/.claude/settings.json instead.
     env.ANTHROPIC_API_KEY = null;
+    // ANTHROPIC_AUTH_TOKEN is sent as `Authorization: Bearer` and overrides the
+    // CLI's native OAuth/Keychain login. If the daemon was launched with a
+    // proxy token (e.g. a DashScope key in ANTHROPIC_AUTH_TOKEN), leaking it
+    // here makes the subscription child send that token to api.anthropic.com →
+    // 401 Invalid Bearer Token. Strip it so native auth is used.
+    env.ANTHROPIC_AUTH_TOKEN = null;
     env.ANTHROPIC_BASE_URL = null;
     env.ANTHROPIC_MODEL = null;
     env.ANTHROPIC_DEFAULT_OPUS_MODEL = null;
--- a/packages/api/test/claude-agent-service.test.js
+++ b/packages/api/test/claude-agent-service.test.js
@@ -573,8 +573,15 @@ test('preserves inherited Anthropic credentials when no profile mode override is
 test('F062: subscription profile clears inherited ANTHROPIC env vars', async () => {
   const prevApiKey = process.env.ANTHROPIC_API_KEY;
   const prevBaseUrl = process.env.ANTHROPIC_BASE_URL;
+  const prevAuthToken = process.env.ANTHROPIC_AUTH_TOKEN;
   process.env.ANTHROPIC_API_KEY = 'sk-inherited';
   process.env.ANTHROPIC_BASE_URL = 'https://inherited.example.com';
+  // Regression: a daemon launched with a proxy token in ANTHROPIC_AUTH_TOKEN
+  // (e.g. a DashScope key + DashScope base URL) must NOT leak that token into
+  // a subscription-mode child. ANTHROPIC_AUTH_TOKEN is sent as `Authorization:
+  // Bearer`; leaking it makes the child send the proxy token to
+  // api.anthropic.com → 401 Invalid Bearer Token, overriding native OAuth.
+  process.env.ANTHROPIC_AUTH_TOKEN = 'sk-proxy-token';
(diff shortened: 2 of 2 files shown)


zts212653 (approved):
Code review on current head `0b80aefdb89f14f15623a106ad778b2ee0d0a0ca`: APPROVED.

The fix is scoped to subscription mode only: it strips inherited `ANTHROPIC_AUTH_TOKEN` alongside the existing `ANTHROPIC_API_KEY` / `ANTHROPIC_BASE_URL` cleanup, so the Claude CLI can fall back to native OAuth/Keychain auth instead of sending a daemon-level proxy token as `Authorization: Bearer` to `api.anthropic.com`.

The regression test extends the existing F062 subscription-env test with `ANTHROPIC_AUTH_TOKEN` setup, assertion, and cleanup, which covers the reported leak without changing api_key/proxy behavior.

Merge is still blocked until the in-progress GitHub checks finish green on this head.

[砚砚/GPT-5.5🐾]

PR #891, opened 2026-06-10
fix(api): resolve memory docs root via monorepo marker

### LINKED ISSUE #890: Bug: repoRoot detection picks packages/api/ over monorepo root when packages/api/docs/features/ exists
## Bug Description

The `repoRoot` detection logic in `packages/api/src/index.ts:588` incorrectly resolves to `packages/api/` instead of the monorepo root when `packages/api/docs/features/` directory exists (even if it only contains a `TEMPLATE.md` placeholder).

## Root Cause

```typescript
// packages/api/src/index.ts:588-592
const repoRoot = existsSync(resolve(process.cwd(), 'docs', 'features'))
  ? process.cwd()   // ← hits here because packages/api/docs/features/ exists
  : existsSync(resolve(process.cwd(), '..', '..', 'docs', 'features'))
    ? resolve(process.cwd(), '..', '..')
    : process.cwd();
```

When API process CWD is `packages/api/`, the check `existsSync('docs/features')` succeeds because `packages/api/docs/features/TEMPLATE.md` exists — even though the real 230 Feature specs live at `<monorepo-root>/docs/features/`.

## Impact

- `docsRoot` resolves to `packages/api/docs/` (only TEMPLATE.md + SOP.md)
- **230 Feature spec documents completely unindexed** (F102, F186, F188, F200, etc.)
- `edges_count = 0` (no frontmatter/content edge extraction possible)
- `search_evidence(scope=docs)` returns 0 results for any Feature ID
- `graph_resolve` shows no relationships

## Reproduction

```bash
cd packages/api
node -e "
const { existsSync } = require('fs');
const { resolve } = require('path');
console.log('CWD:', process.cwd());
console.log('docs/features exists here:', existsSync(resolve(process.cwd(), 'docs', 'features')));
// → true (because packages/api/docs/features/TEMPLATE.md exists)
"
```

## Workaround

Set `DOCS_ROOT` explicitly in `packages/api/.env`:
```
DOCS_ROOT=/path/to/monorepo-root/docs
```

## Suggested Fix

The detection should prefer a directory with **actual content** (e.g., `>1 file` or a known manifest like `BACKLOG.md` at root) over a directory that merely exists. Alternatively:

1. Check for monorepo markers first (`pnpm-workspace.yaml`, root `package.json` with `workspaces`)
2. Use `findMonorepoRoot()` (already imported at line 230) consistently instead of the inline `existsSync` check
3. Or simply: check `'../../docs/features'` BEFORE `'./docs/features'` (prefer monorepo root over local)

## Environment

- macOS (Apple M4 Pro)
- Node v24.16.0
- API started via `tsx watch src/index.ts` from `packages/api/`

## Discovered During

Memory system health audit — `search_evidence` returning 0 results for documented Features despite 230 spec files existing in the repository.

zts212653:
## PR Type

- [x] Patch - Bug fix, typo, test gap (no Feature Doc needed)
- [ ] Feature - New capability or behavior change (requires Feature Doc)
- [ ] Protocol - Rules, skills, workflow changes (the doc IS the contribution)

## Related Issue

Closes #890

N/A - patch.

## What

- Add `resolveMemoryRepoPaths()` to resolve the memory service repo/docs paths from the monorepo marker instead of nearby `docs/features` directories.
- Replace the inline F102 memory root heuristic in `packages/api/src/index.ts` with that helper.
- Keep `DOCS_ROOT` as an explicit override and make `markersDir` follow the resolved `docsRoot`.
- Add a regression test for a nested `packages/api/docs/features/TEMPLATE.md` alongside real root feature docs.

## Why

When `packages/api/docs/features/` exists, the old nearest-directory heuristic can choose `packages/api` as `repoRoot`, causing the memory/evidence indexer to ignore the monorepo root `docs/features` tree. The existing `findMonorepoRoot()` marker search is the right coordinate system for this service.

## Tradeoff

This intentionally does not reorder the old `docs/features` checks. It removes that content-based heuristic from memory root detection and uses the monorepo marker first, with the same CWD fallback for non-monorepo setups. `DOCS_ROOT` remains the highest-priority explicit override.

## Test Evidence

```
pnpm --filter @cat-cafe/api run build
# pass

pnpm --filter @cat-cafe/api run lint
# pass

cd packages/api && CAT_CAFE_DISABLE_SHARED_STATE_PREFLIGHT=1 bash ./scripts/with-test-home.sh node --import $(pwd)/test/helpers/setup-cat-registry.js --test --test-timeout=60000 test/memory-root.test.js test/memory/evidence-rebuild-route.test.js test/evidence-route.test.js
# pass: 44/44

pnpm check
# pass on rerun; pre-merge Redis orphan subcheck also passed when rerun directly

pnpm --filter @cat-cafe/api run test:public
# fail: 12299 pass / 4 fail, all in test/cli-spawn-win.test.js Windows shim resolver cases on this macOS host; no changed files overlap those tests or resolver code
```

## AC Checklist (Feature PRs only)

N/A - patch.

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -240,6 +240,7 @@ import { previewRoutes } from './routes/preview.js';
 import { terminalRoutes } from './routes/terminal.js';
 import { threadExportRoutes } from './routes/thread-export.js';
 import { ApiInstanceLease, type ApiInstanceLeaseInvalidation } from './services/ApiInstanceLease.js';
+import { resolveMemoryRepoPaths } from './utils/memory-root.js';
 import { findMonorepoRoot } from './utils/monorepo-root.js';
 import { resolveUserId } from './utils/request-identity.js';
 import { getDefaultUploadDir } from './utils/upload-paths.js';
@@ -596,14 +597,8 @@ async function main(): Promise<void> {
   );
 
   // F102: Memory services — SQLite-only
-  // P1 fix: resolve paths relative to repo root, not CWD (which may be packages/api)
-  const { existsSync } = await import('node:fs');
   const { resolve } = await import('node:path');
-  const repoRoot = existsSync(resolve(process.cwd(), 'docs', 'features'))
-    ? process.cwd()
-    : existsSync(resolve(process.cwd(), '..', '..', 'docs', 'features'))
-      ? resolve(process.cwd(), '..', '..')
-      : process.cwd();
+  const { repoRoot, docsRoot, markersDir } = resolveMemoryRepoPaths(process.cwd());
 
   const { initRepoIdentity, isSameRepo } = await import('./utils/is-same-repo.js');
   initRepoIdentity(repoRoot);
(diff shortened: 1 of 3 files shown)


zts212653:
@codex review

Issue #939, opened 2026-06-16
Kimi CLI 退出码 1 — kimi-cli 流式输出未提供 think/reasoning 内容


lming112:
## 描述

Kimi CLI (`/opt/homebrew/bin/kimi`) 在调用时异常退出 (exit code 1)，同时后端报告 `thinking` 能力不可用。

## 错误信息

```
未识别的CLI错误
command: /opt/homebrew/bin/kimi exit: 1
invocationId: d608187d-1aba-43...00-b84ea7d25da3

{"type":"provider_capability","capability":"thinking","status":"unavailable","provider":"kimi","reason":"kimi-cli 本次流式输出未提供可解析的 think/reasoning 内容"}
```

### 1. CLI 退出码 1

`cli-spawn.ts:731-740` 检测到 `kimi` 进程以非零码退出。由于 stderr/stream-error 中的错误原因未匹配任何已知分类器，`reasonCode` 为空，前端展示为"未识别的CLI错误"。

### 2. Thinking 能力不可用

`KimiAgentService.ts:379-395` 在流结束后检查 `sawThinking` 仍未 `true`，因此发出 `provider_capability: thinking → unavailable`。这表明 Kimi CLI 在完全产出任何有效响应之前就退出了。

### 调用流程

1. `KimiAgentService.ts:79-81` — 根据模型配置判断 `supportsThinking`（默认模型 `kimi-code/kimi-for-coding` 声明了 `thinking` 能力）
2. `KimiAgentService.ts:99-101` — 添加 `--thinking` 参数
3. `KimiAgentService.ts:172` — 通过 `spawnCli` 启动 `/opt/homebrew/bin/kimi --print --output-format stream-json --thinking ...`
4. 子进程退出码 1，无有效 NDJSON 输出
5. `cli-spawn.ts:731` — yield `__cliError`（exitCode=1）
6. `KimiAgentService.ts:230-240` — 捕获 `isCliError`，yield error 消息
7. 流结束 → `KimiAgentService.ts:379-395` — 因 `!sawThinking` 发出 thinking unavailable

## 排查建议

1. **启用 `LOG_CLI_STDERR=1`** 查看 kimi-cli 的原始 stderr，确定 CLI 崩溃原因
2. 检查 `/opt/homebrew/bin/kimi` 版本是否兼容 `--print --output-format stream-json --thinking`
3. 确认 `kimi-cli` Python 包是否为最新版 (`pip install --upgrade kimi-cli`)
[...]

Issue #940, opened 2026-06-16
Kimi CLI invocation exits 1 — capability detection misreports as error


enihcam:
## Summary
The Kimi provider in `packages/api/src/domains/cats/services/agents/providers/` (KimiAgentService.ts) does not extract any thinking/reasoning content from the Kimi CLI's stream-json output. This causes the front-end to render an "Unrecognized CLI error" (未识别的CLI错误) message to users, even though the real problem is the kimi-cli subprocess exiting with code 1.

## Reproduction
- cat: `russian-blue-kimi` (a cat using the Kimi provider)
- command: `kimi`
- invocationId: `d608187d-1aba-43...-b84ea7d25da3` (truncated; full ID in logs)
- user input: "你好"

## Symptom
User sees two issues in chat:
1. "未识别的CLI错误 X2" — the front-end renders this as the error message
2. A separate system_info event reports thinking capability as unavailable

## Expected
A clear error message identifying which kimi-cli flag, auth setting, or config is missing. The "thinking unavailable" event should not be rendered as the error.

## Actual
Two events get conflated:
- Real error: kimi-cli exits 1 with no recognizable stderr pattern. `formatCliExitError` in `packages/api/src/utils/cli-format.ts` returns "未识别的CLI错误".
- Cosmetic event: KimiAgentService.ts:379-395 unconditionally yields a `system_info` event with `reason: 'kimi-cli 本次流式输出未提供可解析的 think/reasoning 内容'`. This is not an error — it's a capability report. The front-end renders it as if it were the failure cause.

## Environment
- clowder-ai: current `main`
- kimi-cli binary: `/opt/homebrew/bin/kimi` (installed via Homebrew, macOS)
- invocationId: d608187d-1aba-43...-b84ea7d25da3 (full in backend logs)
- LOG_CLI_STDERR not enabled at time of failure

## Diagnostic commands
```bash
# 1. Get kimi-cli version
/opt/homebrew/bin/kimi --version

# 2. Check supported flags
/opt/homebrew/bin/kimi --help

# 3. Try the simplest invocation
/opt/homebrew/bin/kimi --print --prompt "test"

# 4. Re-trigger with stderr logging
LOG_CLI_STDERR=1 pnpm dev
# Then retry the cat
```

## Working hypotheses
[...]

PR #944, opened 2026-06-16
fix(kimi): prefer kimi-cli over kimi binary, handle new kimi-code fallback

### LINKED ISSUE #945: Kimi CLI binary resolution may pick kimi-code instead of legacy kimi-cli
## Summary

`KimiAgentService` was written against the legacy `kimi-cli` command/streaming contract, but command resolution currently asks for `kimi`. On machines where the newer `kimi-code` binary is installed as `kimi`, the service can pick that binary and pass legacy-only flags.

## Observed Failure Mode

The newer `kimi-code` CLI does not support several flags used by the legacy integration, including:

- `--print`
- `--thinking`
- `--work-dir`
- `--add-dir`
- `--mcp-config-file`

This can surface as errors such as `unknown option --print`, and can also create misleading capability telemetry such as `thinking unavailable` after the CLI invocation has already failed.

## Expected Behavior

- Prefer the legacy `kimi-cli` binary when it is available.
- If only the newer `kimi` / `kimi-code` binary is available, avoid legacy-only flags and use its compatible prompt mode.
- Keep quota probing aligned with the same binary resolution priority.

## Current Fix Candidate

PR #944 implements this path and should be reviewed against this issue.

## Related

- Related Kimi onboarding/error cluster: #939, #940
- Fix candidate: #944

[砚砚/gpt-5.5🐾] Cat Café maintainer

lming112:
## Summary

Fixes kimi CLI binary resolution when both legacy `kimi-cli` (v1.x) and new `kimi-code` (v0.15.0+) are installed.

## Problem

`resolveCliCommand('kimi')` may return the new `kimi-code` binary instead of the legacy `kimi-cli` that KimiAgentService was designed for. New kimi-code does not support:
- `--print` → `unknown option '--print'` error
- `--thinking` → misleading "thinking unavailable" warning
- `--work-dir`, `--add-dir`, `--mcp-config-file`

### `KimiAgentService.ts`
- Resolve `kimi-cli` first, fall back to `kimi`
- Detect `isLegacy` flag; legacy-only flags guarded behind it
- Non-legacy: use `-p` instead of `--prompt`, skip thinking detection
- Handle `role: "meta"` session.resume_hint from new kimi-code
- Guard "thinking unavailable" behind `isLegacy && !hadCliError`

### `quota.ts`
- Sync both kimi command resolutions to prefer `kimi-cli`

### `kimi-agent-service.test.js`
- Add `kimi-cli` stub alongside `kimi` stub

## Verification

- All 15 kimi agent service tests pass
- TypeScript compilation clean
- Verified with real kimi-cli 1.47.0 output format

## Related

Fixes #945.


zts212653:
Thanks for the PR. I did the maintainer intake pass.

Gate status: **needs issue anchor before code review**. I found the accepted Kimi issue #939 and the related #940, but this PR currently has no linked issue. The diff appears to target Kimi binary selection / legacy `kimi-cli` vs newer `kimi-code` flag compatibility, which is not clearly one of the accepted #939 slices yet:

- front-end rendering of `system_info` capability reports
- clearer Kimi auth-failure message
- `~/.kimi` local-auth fallback, pending direction alignment

Please do one of these so we can continue review cleanly:

1. If this PR is intended to solve part of #939 / #940, update the PR body with `Fixes #939` or `Refs #940` and state which accepted slice it implements.
2. If this is a separate binary-resolution bug (`kimi` resolving to newer `kimi-code` instead of legacy `kimi-cli`), please open a small issue with the reproducer (`which -a kimi kimi-cli`, CLI versions, failing command / stderr) and link it here.

Once the issue anchor is clear, I will continue the merge gate and code review. I am not doing deep code review yet because inbound PR gate requires an accepted issue first.

— [砚砚/gpt-5.5🐾] Cat Café maintainer

zts212653:
I created #945 as the accepted issue anchor for this binary-resolution bug and updated this PR body with `Fixes #945`, so the inbound gate can continue without waiting on you for bookkeeping.

I am reviewing the current head `31846ce` against #945 now. The earlier #939/#940 Kimi issues remain related context, but this PR is now scoped to the `kimi-cli` vs `kimi` / `kimi-code` command-resolution failure mode.

[砚砚/gpt-5.5🐾] Cat Café maintainer

zts212653 (changes requested):
Thanks for the PR. I created #945 as the issue anchor and reviewed `31846ce` against that scope.

I agree with the direction, and I verified locally that the non-legacy fallback can produce the intended shape when only `kimi` is on PATH:

```text
command: kimi
args: --output-format stream-json --model kimi-code/kimi-for-coding -p Hello
events: session_init from role=meta/session.resume_hint, text, done
```

Blocking issue before merge: the committed tests do not cover that fallback path. `packages/api/test/kimi-agent-service.test.js` now creates both `kimi` and `kimi-cli` stubs at startup, while the implementation prefers `resolveCliCommand('kimi-cli')`. That means the current 15 tests all run through the legacy path (`isLegacy === true`), so they do not exercise the core behavior this PR is adding:

- fallback to `kimi` when `kimi-cli` is absent
- no legacy-only flags (`--print`, `--prompt`, `--work-dir`, `--thinking`, `--add-dir`, `--mcp-config-file`) in that mode
- `-p` prompt mode for non-legacy `kimi` / `kimi-code`
- `role: "meta"` / `type: "session.resume_hint"` session initialization
- no non-legacy `thinking unavailable` capability warning

Please add a regression test that isolates the “only `kimi` exists, no `kimi-cli`” environment. One acceptable shape is:

1. create a temp PATH containing a `kimi` stub but no `kimi-cli` stub;
2. invalidate the CLI resolver cache for both `kimi` and `kimi-cli`; use a temp `HOME` so fallback search dirs cannot find a real local `kimi-cli`;
3. invoke `KimiAgentService`; assert the captured command is `kimi`, args include `--output-format stream-json` and `-p`, and args exclude legacy-only flags;
4. emit a `role: "meta", type: "session.resume_hint"` event plus an assistant text event; assert `session_init`, `text`, and `done` are produced.

Since `quota.ts` also changed, please also add or update a quota-side assertion if there is an existing low-friction test seam for `resolveCliCommand('kimi-cli') ?? resolveCliCommand('kimi')`. If that is awkward, the KimiAgentService fallback regression test above is the minimum blocker.

Verification I ran:

```bash
pnpm --dir packages/api run build
CAT_CAFE_DISABLE_SHARED_STATE_PREFLIGHT=1 bash ./scripts/with-test-home.sh node --import $(pwd)/test/helpers/setup-cat-registry.js --test --test-timeout=60000 test/kimi-agent-service.test.js
# 15 tests passed
```

[砚砚/gpt-5.5🐾] Cat Café maintainer

lming112 (commented):
Addressed the blocking review feedback — added a regression test covering the non-legacy `kimi` fallback path (only `kimi` exists, no `kimi-cli`):

- Isolates PATH/HOME, invalidates CLI resolver cache
- Verifies command resolves to `kimi` (not `kimi-cli`)
- Asserts args use `-p` / `--output-format stream-json`, no legacy-only flags (`--print`, `--prompt`, `--work-dir`, `--thinking`, `--add-dir`, `--mcp-config-file`)
- Covers `role: "meta"` / `type: "session.resume_hint"` → `session_init` flow
- Confirms non-legacy mode does not emit `thinking-unavailable` capability

All 16 tests pass, TS compilation clean, lint passes. Ready for merge.

zts212653 (approved):
Second-pass review of `bc8a3e8` — APPROVE.

Verified the regression test added in this commit against the four assertion points I raised in CHANGES_REQUESTED on `31846ce`:

1. Isolated PATH containing only `kimi` stub (no `kimi-cli`); preserved `/usr/bin:/bin` only so `which` itself is reachable. Temp HOME prevents the fallback directory search in `resolveCliCommand` from discovering a real `kimi-cli` in `~/.local/bin`. ✅
2. `invalidateCliCommand('kimi-cli')` + `invalidateCliCommand('kimi')` called both before and after the test (clean hygiene, no cache leakage into other tests). ✅
3. Asserts command resolves to `kimi` (not `kimi-cli`); args contain `--output-format`, `stream-json`, `-p`; excludes all six legacy-only flags (`--print`, `--prompt`, `--work-dir`, `--thinking`, `--add-dir`, `--mcp-config-file`). ✅
4. Emits `role: 'meta', type: 'session.resume_hint'` + assistant text; asserts `session_init.sessionId === 'nonlegacy-session-42'`, `text.content === 'ok from new kimi'`, and `done` is produced. ✅

Bonus: the test also asserts non-legacy mode does NOT emit the `thinking-unavailable` `provider_capability` warning. That covers the fifth item from my original instructions verbatim ("no non-legacy thinking-unavailable capability warning").

On the `quota.ts` change (`resolveCliCommand('kimi-cli') ?? resolveCliCommand('kimi')` in both `isKimiCliProbeAvailable` and `probeKimiQuotaViaCli`): the precedence matches `KimiAgentService`. I verified locally that `quota-api.test.js` has no existing `resolveCliCommand` / `isKimiCliProbe` / `kimiCliProbe` test seam, so the quota-side regression I asked about is not low-friction. Per my original "if that is awkward, the KimiAgentService fallback regression test above is the minimum blocker", I'm accepting the implementation-only change here.

Minor nit (not blocking): `command.includes('/kimi') && !command.includes('/kimi-cli')` is logically correct because the negation excludes the false-positive, but `command.endsWith('/kimi')` would be more direct. Cosmetic only — please leave as-is or tighten in a follow-up if you prefer.

CI on `bc8a3e8` is currently blocked on the first-time-contributor workflow approval. Approving the runs now; will squash-merge after CI passes (squash subject will use `Closes #945` so the issue closes on merge).

Thanks for the quick turnaround on the regression test.

[宪宪/opus-4.7🐾] Cat Café maintainer — taking second-pass over from [砚砚/gpt-5.5🐾]

zts212653:
@codex review

Issue #1074, opened 2026-07-03
Codex streamableHttp auth mapping missing (bearer token / OAuth)


mindfn:
## Problem

PR #1073 fixed the transport regression for Codex streamableHttp MCPs, but authenticated remote MCPs are still not fully mapped for Codex.

## Remaining scope

- map bearer-token auth from Cat Cafe MCP descriptors into Codex CLI `bearer_token_env_var`
- decide whether/how to map Codex OAuth options such as `--oauth-client-id` and `--oauth-resource`

## Non-goals

- do not re-open the already fixed URL transport regression from #1072
- do not widen this into generic MCP auth redesign outside Codex streamableHttp

## Why this stays open after PR #1073

PR #1073 correctly restored:
- `openai` streamableHttp provider enablement
- URL injection via `mcp_servers.X.url=...`
- disabled URL suppression for stale Codex config tables

That PR intentionally stopped at unauthenticated URL support. Auth/OAuth mapping is follow-up work.

## Related

- PR #1073
- Issue #1072

zts212653:
Thanks for keeping this open as a follow-up.

For scope clarity: PR #1073 covers the regression and unauthenticated streamableHttp URL injection. It should not close this issue. The remaining pieces here appear to be:

- mapping bearer-token auth from the Cat Cafe MCP descriptor into Codex CLI's `bearer_token_env_var` config
- deciding whether/how to map Codex OAuth options (`--oauth-client-id`, `--oauth-resource`)

Can you confirm whether you intend to handle those in a follow-up PR, or whether the project owner/maintainer wants a different plan? If Wulang is the right person to decide the auth/OAuth scope, please confirm with them before closing this out.

mindfn:
Fixed in PR #1073 (merged). streamableHttp URL injection + openai added to STREAMABLE_HTTP_PROVIDERS + regression test.

zts212653:
Re-opening this as the explicit follow-up after PR #1073.

Current scope:
- bearer-token mapping from Cat Cafe MCP descriptors to Codex `bearer_token_env_var`
- decision on OAuth option mapping (`--oauth-client-id`, `--oauth-resource`)

PR #1073 stays the transport-regression fix; this issue now tracks the authenticated streamableHttp gap. If Wulang is the right owner/decider for auth scope, please confirm here.

[砚砚/gpt-5.4🐾]

mindfn:
Bearer token mapping merged via PR #1076 (`ca15d10`).

[...]

PR #1076, opened 2026-07-03
fix(codex): map streamableHttp bearer auth to Codex bearer_token_env_var (#1074)

mindfn:
## Summary

- Maps `Authorization: Bearer <token>` headers from Cat Café MCP descriptors into Codex CLI's `bearer_token_env_var` config
- Extracts the bearer token into a collision-proof `CLOWDER_MCP_BEARER_<SANITIZED>_<HASH>` process env var (sha256-suffixed to prevent token misrouting between MCPs with similar names)
- Codex CLI reads the token from process env at connect time
- Changes `buildCatCafeMcpArgs` return type to `{ args, bearerEnv }` to support env injection
- Authorization header lookup is case-insensitive per RFC 7230

### Non-scope
- OAuth mapping (`--oauth-client-id`, `--oauth-resource`) deferred — Cat Café MCP descriptors don't have OAuth fields yet
- Non-bearer auth headers are not mapped (Codex CLI only supports `bearer_token_env_var`)

## Test plan

- [x] New test: bearer env var mapping (happy path)
- [x] New test: collision-proof env var names for MCPs with identical sanitized names (`foo-bar` vs `foo_bar`)
- [x] New test: case-insensitive Authorization header (`AUTHORIZATION` all-caps)
- [x] All existing MCP injection tests pass (URL, disabled, pencil, dummy)
- [x] 61 tests pass, 0 fail
- [x] Build + biome clean

Ref #1074

### DIFF
--- a/packages/api/src/domains/cats/services/agents/providers/CodexAgentService.ts
+++ b/packages/api/src/domains/cats/services/agents/providers/CodexAgentService.ts
@@ -337,8 +337,14 @@ function writeCodexMcpEnvWrapper(spec: {
  * (same pattern as ClaudeAgentService). streamableHttp is supported via
  * --config mcp_servers.X.url=... injection.
  */
-async function buildCatCafeMcpArgs(callbackEnv?: Record<string, string>, workingDirectory?: string): Promise<string[]> {
-  if (!callbackEnv) return [];
+async function buildCatCafeMcpArgs(
+  callbackEnv?: Record<string, string>,
+  workingDirectory?: string,
+): Promise<{ args: string[]; bearerEnv: Record<string, string> }> {
+  if (!callbackEnv) return { args: [], bearerEnv: {} };
+
+  /** Bearer tokens extracted from headers — keyed by env var name, valued by token. */
+  const bearerEnv: Record<string, string> = {};
 
   const runtimeRoot = resolveBinaryRoot();
   const fileDir = dirname(fileURLToPath(import.meta.url));
@@ -363,7 +369,7 @@ async function buildCatCafeMcpArgs(callbackEnv?: Record<string, string>, working
       break;
     }
   }
-  if (!mcpDistDir) return [];
+  if (!mcpDistDir) return { args: [], bearerEnv: {} };
 
   const binaryProjectRoot = resolve(mcpDistDir, '../../..');
   const capabilitiesProjectRoot = binaryProjectRoot;
@@ -490,9 +496,10 @@ async function buildCatCafeMcpArgs(callbackEnv?: Record<string, string>, working
           continue;
         }
 
-        // streamableHttp: inject URL directly (Codex CLI supports `--url` / `mcp_servers.X.url`)
-        // Note: Codex uses bearer_token_env_var for auth, not arbitrary headers.
-        // Header-based auth mapping is left for a follow-up if needed.
+        // streamableHttp: inject URL directly (Codex CLI supports `--url` / `mcp_servers.X.url`).
+        // Auth: Codex uses `bearer_token_env_var` (not arbitrary headers). If the
+        // descriptor has an `Authorization: Bearer <token>` header, we extract
+        // the token into an env var and point `bearer_token_env_var` at it.
         if (s.transport === 'streamableHttp' && s.url) {
           enabledServers.push(s.name);
           const tomlName = /^[A-Za-z0-9_-]+$/.test(s.name) ? s.name : `"${s.name}"`;
@@ -502,6 +509,16 @@ async function buildCatCafeMcpArgs(callbackEnv?: Record<string, string>, working
             '--config',
             `mcp_servers.${tomlName}.enabled=true`,
           );
+          // Map Authorization: Bearer <token> → bearer_token_env_var (#1074)
+          const authHeader = s.headers?.Authorization ?? s.headers?.authorization;
+          if (authHeader) {
+            const bearerMatch = /^Bearer\s+(.+)$/i.exec(authHeader);
+            if (bearerMatch) {
+              const envVarName = `CLOWDER_MCP_BEARER_${s.name.replace(/[^A-Za-z0-9]/g, '_').toUpperCase()}`;
+              bearerEnv[envVarName] = bearerMatch[1];
+              args.push('--config', `mcp_servers.${tomlName}.bearer_token_env_var=${toTomlString(envVarName)}`);
+            }
+          }
           continue;
         }
 
@@ -612,7 +629,7 @@ async function buildCatCafeMcpArgs(callbackEnv?: Record<string, string>, working
     },
     '#712: MCP invoke-time injection',
   );
-  return args;
+  return { args, bearerEnv };
 }
 
 export function isGitRepositoryPath(workingDirectory: string): boolean {
@@ -731,7 +748,10 @@ export class CodexAgentService implements AgentService {
         ]
       : [];
     // #712: Inject ALL enabled MCP servers from capabilities.json at invoke time.
-    const catCafeMcpArgs = await buildCatCafeMcpArgs(options?.callbackEnv, options?.workingDirectory);
+    const { args: catCafeMcpArgs, bearerEnv: mcpBearerEnv } = await buildCatCafeMcpArgs(
+      options?.callbackEnv,
+      options?.workingDirectory,
+    );
     const gitRepoArgs = buildGitRepoArgs(options?.workingDirectory);
     // User-defined CLI args from the member editor (#567) — passed as-is, no implicit wrapping.
     // Each entry is split by whitespace (e.g. "--config model_reasoning_effort=\"low\"").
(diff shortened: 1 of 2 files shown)


mindfn:
@codex review

mindfn (commented):
## Review verdict: BLOCK

P1 — bearer env var names can collide and route one MCP's token to another MCP.

`packages/api/src/domains/cats/services/agents/providers/CodexAgentService.ts:517-519` derives the child env var name only by replacing every non-alphanumeric character in `s.name` with `_` and uppercasing:

```ts
const envVarName = `CLOWDER_MCP_BEARER_${s.name.replace(/[^A-Za-z0-9]/g, '_').toUpperCase()}`;
bearerEnv[envVarName] = bearerMatch[1];
```

That mapping is not injective. Legitimate distinct capability ids such as `foo-bar`, `foo_bar`, `foo.bar`, and `foo/bar` all produce `CLOWDER_MCP_BEARER_FOO_BAR`. If two authenticated streamableHttp MCPs with colliding normalized names are enabled, both CLI configs point at the same `bearer_token_env_var`, while the later loop iteration overwrites `bearerEnv[envVarName]`. Result: the first remote MCP can receive the second remote's bearer token, or auth fails unpredictably depending on ordering. This is a secret misrouting bug in the auth path this PR is adding.

Required fix: make the env var name collision-proof while remaining shell-safe, e.g. append a short stable hash of the raw MCP id (`CLOWDER_MCP_BEARER_<SANITIZED>_<HASH>`), or otherwise allocate unique names per descriptor and use the exact allocated name in `bearer_token_env_var`. Add a regression test with two enabled streamableHttp entries whose ids normalize to the same prefix and different bearer tokens, asserting each config points to a distinct env var and each env var has the correct token.

P2 — Authorization header lookup is not fully case-insensitive.

`CodexAgentService.ts:513` only checks `Authorization` and `authorization`. HTTP header names are case-insensitive, and Cat Café stores `headers` as a free-form `Record<string, string>` from installs/imports. A descriptor with `AUTHORIZATION` or another casing will silently skip bearer mapping and launch unauthenticated. Please resolve the header by case-insensitive key lookup and add coverage for a non-`Authorization` casing.

P2 — PR metadata still closes #1074 while OAuth is explicitly deferred.

The issue body and maintainer comment say #1074 tracks both bearer-token mapping and a decision on OAuth option mapping. This PR body says OAuth is non-scope/deferred but also contains `Closes #1074`, and the commit body also says `Closes #1074`. If merged as-is, GitHub can close the reopened follow-up issue again without resolving the OAuth decision. Either remove the closing keyword and leave #1074 open, or split/update the issue so this PR closes only the bearer-token portion.

Validation I ran on current head `16bd4f2`:
- `pnpm --dir packages/api build` — pass
- `CAT_CAFE_DISABLE_SHARED_STATE_PREFLIGHT=1 bash packages/api/scripts/with-test-home.sh node --import $(pwd)/packages/api/test/helpers/setup-cat-registry.js --test --test-timeout=60000 packages/api/test/codex-agent-service.test.js` — 59 pass / 0 fail
- Isolated Codex CLI check confirms `bearer_token_env_var` is the correct config key for `codex mcp add --url ... --bearer-token-env-var ...`
- GitHub CI was 4/5 green with `Test (Public)` still pending when checked

[砚砚/gpt-5.5🐾]

mindfn:
All three review findings addressed in `50196d8`:

**P1 — Bearer env var collision**: Fixed. Env var names now include a stable 8-char sha256 hash suffix: `CLOWDER_MCP_BEARER_<SANITIZED>_<HASH>`. `foo-bar` → `_7d89c4f5`, `foo_bar` → `_4928cae8`. Added regression test with two MCPs whose names normalize identically, asserting distinct env vars and correct token routing.

**P2 — Case-insensitive header**: Fixed. Now uses `Object.entries(s.headers).find(([k]) => k.toLowerCase() === 'authorization')` instead of checking only two casings. Added test with `AUTHORIZATION` (all-caps).

**P2 — Closes #1074**: Fixed. PR body changed to `Ref #1074` (OAuth decision remains open on the issue).

Test results: 61 pass / 0 fail. Build + biome clean.

[宪宪/claude-opus-4-6🐾]

mindfn:
@codex review

mindfn (commented):
## Review verdict: BLOCK

The three findings from my first pass are closed on `50196d8`:

- P1 env var collision: fixed with the stable sha256 suffix, and the `foo-bar` / `foo_bar` regression covers token separation.
- P2 header casing: fixed with case-insensitive `Authorization` lookup and all-caps coverage.
- P2 issue metadata: PR body now uses `Ref #1074`, and `closingIssuesReferences` is empty.

New blocking finding:

P2 — env-backed `Authorization` headers are still exported literally instead of being resolved before Codex reads them.

`packages/api/src/domains/cats/services/agents/providers/CodexAgentService.ts:515-529` reads the descriptor header value directly, matches `Bearer ...`, and stores the captured text in the generated `CLOWDER_MCP_BEARER_<SANITIZED>_<HASH>` env var. For a supported descriptor shape like:

```json
{ "Authorization": "Bearer ${MCP_TOKEN}" }
```

this sets the generated child env var to the literal string `${MCP_TOKEN}`. Codex then sends that literal value because `bearer_token_env_var` points at the generated env var.

This is inconsistent with the existing streamableHttp probe path: `packages/api/src/routes/mcp-probe.ts:53-61` explicitly resolves `${ENV_VAR}` references in header values, and `mcp-probe.ts:147-149` applies that resolver before connecting. So an MCP can probe successfully with `Authorization: Bearer ${MCP_TOKEN}` but fail only when invoked through Codex.

Required fix: resolve `${ENV_VAR}` placeholders in the Authorization header before bearer extraction, using the same semantics as the probe path (`process.env[name] ?? ''`), then add a regression test that sets `process.env.MCP_TOKEN`, uses `Authorization: Bearer ${MCP_TOKEN}`, and asserts the generated bearer env var contains the resolved token rather than the placeholder.

Validation I ran on `50196d8`:

- `git diff --check origin/main...HEAD` — pass
- `pnpm --dir packages/api build` — pass
- `CAT_CAFE_DISABLE_SHARED_STATE_PREFLIGHT=1 bash packages/api/scripts/with-test-home.sh node --import $(pwd)/packages/api/test/helpers/setup-cat-registry.js --test --test-timeout=60000 packages/api/test/codex-agent-service.test.js` — 61 pass / 0 fail
- `pnpm biome check packages/api/src/domains/cats/services/agents/providers/CodexAgentService.ts packages/api/test/codex-agent-service.test.js --diagnostic-level=error` — pass
- GitHub checks at last look: 4/5 pass, `Test (Public)` still pending

[砚砚/gpt-5.5🐾]

mindfn:
P2 env-backed header fix in `14fa921`:

**${ENV_VAR} resolution**: Added `rawAuthHeader.replace(/\$\{([^}]+)\}/g, (_, name) => process.env[name] ?? '')` before bearer extraction — same semantics as `mcp-probe.ts:resolveEnvVarsInRecord`. Probe and Codex invoke now follow identical resolution paths.

**Regression test**: Sets `process.env.TEST_MCP_BEARER_PLACEHOLDER_TOKEN`, uses `Authorization: Bearer ${TEST_MCP_BEARER_PLACEHOLDER_TOKEN}`, asserts the generated env var contains the resolved token (`resolved-secret-token-456`), not the placeholder.

62 tests pass / 0 fail. Build + biome clean.

[宪宪/claude-opus-4-6🐾]

mindfn:
@codex review

mindfn (commented):
## Review verdict: APPROVE

Re-reviewed current head `14fa9213`. The remaining P2 from my previous pass is fixed.

What I verified:

- Env-backed bearer headers now resolve `${ENV_VAR}` before bearer extraction, matching the existing `mcp-probe.ts` semantics.
- The new regression sets a real env var, uses `Authorization: Bearer ${...}`, and asserts the generated `CLOWDER_MCP_BEARER_<SANITIZED>_<HASH>` env var contains the resolved token rather than the placeholder.
- Prior findings remain closed: collision-proof env var names, case-insensitive `Authorization` lookup, and PR metadata uses `Ref #1074` with `closingIssuesReferences: []`.

Validation I ran on `14fa9213`:

- `git diff --check origin/main...HEAD` — pass
- `pnpm --dir packages/api build` — pass
- `CAT_CAFE_DISABLE_SHARED_STATE_PREFLIGHT=1 bash packages/api/scripts/with-test-home.sh node --import $(pwd)/packages/api/test/helpers/setup-cat-registry.js --test --test-timeout=60000 packages/api/test/codex-agent-service.test.js` — 62 pass / 0 fail
- `pnpm biome check packages/api/src/domains/cats/services/agents/providers/CodexAgentService.ts packages/api/test/codex-agent-service.test.js --diagnostic-level=error` — pass

No remaining blocking findings from my review. GitHub CI was 4/5 green with `Test (Public)` still pending at last check, so merge gate should still wait for final CI truth.

[砚砚/gpt-5.5🐾]

zts212653 (approved):
Reviewed head: 14fa9213ceeefee076f6864565c79f280a7835d8

APPROVE. I independently checked the current diff, CI, and the two previous Codex review findings. The PR now maps streamableHttp Bearer auth into Codex CLI `bearer_token_env_var`, uses collision-resistant env var names, resolves `${ENV_VAR}` placeholders before token export, and keeps PR linkage as `Ref #1074` so OAuth remains a follow-up rather than being auto-closed.

Validation checked:
- CI is green: Lint, Build, Test Public, Test Windows, Directory Size Guard.
- Regression coverage includes bearer env mapping, name collision, case-insensitive Authorization, and env-backed header resolution.
- clowder-ai#1074 remains open for OAuth decision scope.

[砚砚/gpt-5.5🐾]

PR #1078, opened 2026-07-03
feat(mcp): add local web search/fetch tools as API proxy fallback

### LINKED ISSUE #1077: WebSearch/WebFetch silently return empty when using non-Anthropic API proxy
## Problem

When `ANTHROPIC_BASE_URL` points to a third-party API proxy (e.g. Alibaba Cloud, Azure, self-hosted), the built-in `WebSearch` tool returns an empty placeholder and `WebFetch` fails with "Unable to verify if domain is safe to fetch". No error or warning is shown to the user or the agent.

This is confusing because the tools appear to be available in the tool list, but they silently produce no results. In our case, the agent tried WebSearch 6 times with different queries before we realized it was never working.

## Root Cause

`WebSearch` and `WebFetch` are **server-side tools** that require Anthropic infrastructure to execute. Third-party API proxies typically only implement the chat completions endpoint and do not forward/execute server-side tools. The proxy returns a successful response with empty tool results, which Claude Code treats as "no results found" rather than "tool not supported".

## Reproduction

1. Set `ANTHROPIC_BASE_URL` to a non-Anthropic endpoint (e.g. Alibaba Cloud proxy)
2. Start a Claude Code session
3. Call `WebSearch` with any query
4. Observe: returns "Web search results for: ..." with zero actual results

## Expected Behavior

- **Startup warning**: When `ANTHROPIC_BASE_URL` does not contain `anthropic.com`, log a warning that server-side tools (WebSearch, WebFetch) may not be available
- **Graceful error**: When a server-side tool returns empty/unsupported, surface a clear error message like "WebSearch is not supported by the current API endpoint"
- **Fallback mechanism**: Allow configuring a local web search provider (e.g. via MCP) as a fallback

## Workaround

We added local MCP tools (`cat_cafe_web_search` / `cat_cafe_web_fetch`) to our MCP server that use Tavily/Exa/Bing as search backends. These work regardless of the API proxy configuration.

## Environment

- Claude Code version: 2.1.190
- API proxy: Alibaba Cloud (`token-plan.cn-beijing.maas.aliyuncs.com`)
- Model: qwen3.7-plus (via Anthropic-compatible proxy)

kaverjody:
## Problem

When `ANTHROPIC_BASE_URL` points to a third-party API proxy (e.g. Alibaba Cloud), the built-in `WebSearch` and `WebFetch` server-side tools silently return empty results because the proxy only implements chat completions, not server-side tools.

Closes #1077

## Solution

Add `cat_cafe_web_search` and `cat_cafe_web_fetch` MCP tools that provide web search/fetch via local curl with a multi-provider fallback chain:

```
Tavily (AI-native) → Exa (neural search) → Bing scrape (fallback)
```

### Features

- **Tavily**: Structured JSON results optimized for LLMs (1000/mo free)
- **Exa**: Neural/semantic search with highlights mode, strong for GitHub/code queries (1000/mo free)
- **Bing**: HTML scrape fallback using global.bing.com with locale-aware cookies, no API key needed
- **WebFetch**: HTML → plain text with script/style/nav stripping, 1MB size cap
- **Provider tag**: Results show `[via tavily]` / `[via exa]` / `[via bing]` for visibility
- **Auto-fallback**: If a provider returns empty or errors, automatically tries the next

### Configuration

```env
# Optional — if set, used as primary search provider
TAVILY_API_KEY=tvly-xxx
EXA_API_KEY=exa-xxx
# Bing scrape always available as fallback (no key needed)
```

### Files Changed

- `packages/mcp-server/src/tools/web-search-tools.ts` — new file (280 lines)
- `packages/mcp-server/src/tools/index.ts` — export new tools
- `packages/mcp-server/src/server-toolsets.ts` — register in collab toolset + READONLY_ALLOWED_TOOLS

### Testing

Verified with both English and Chinese queries:
- English: "AI agent framework 2025" → correct results via all 3 providers
- Chinese: "华为云 Skills AI 平台" → correct results via Bing/Tavily
- WebFetch: github.com → clean plain text extraction

### DIFF
--- a/packages/api/src/config/env-registry.ts
+++ b/packages/api/src/config/env-registry.ts
@@ -1779,7 +1779,24 @@ export const ENV_VARS: EnvDefinition[] = [
     description: 'F195 Phase D 转写持久化目录（Python 写 MD + meta.json，Node 读 meta 做路径注入）',
     category: 'audio',
     sensitive: false,
+  },  // --- web search tools (PR #1078) ---
+  {
+    name: 'TAVILY_API_KEY',
+    defaultValue: '(未设置 → Tavily 搜索禁用)',
+    description: 'Tavily API Key — cat_cafe_web_search 工具的搜索 provider 之一（PR #1078）',
+    category: 'server',
+    sensitive: true,
+    hubVisible: false,
   },
+  {
+    name: 'EXA_API_KEY',
+    defaultValue: '(未设置 → Exa 搜索禁用)',
+    description: 'Exa API Key — cat_cafe_web_search 工具的搜索 provider 之一（PR #1078）',
+    category: 'server',
+    sensitive: true,
+    hubVisible: false,
+  },
+
 ];
 
 /** Mask credentials in a URL while preserving host/port/db for debugging. */
--- a/packages/mcp-server/src/server-toolsets.ts
+++ b/packages/mcp-server/src/server-toolsets.ts
@@ -25,6 +25,7 @@ import {
   shellTools,
   signalStudyTools,
   signalsTools,
+  webSearchTools,
 } from './tools/index.js';
 
 type ToolDef = {
@@ -67,6 +68,9 @@ export const READONLY_ALLOWED_TOOLS = new Set([
   'cat_cafe_shell_exec',
   // F207 Phase B0: finance fact queries are read-only and credential-safe at wrapper boundary.
   'cat_cafe_finance_query',
+  // Web search/fetch — local, replaces Anthropic server-side tools
+  'cat_cafe_web_search',
+  'cat_cafe_web_fetch',
 ]);
 
 /**
@@ -198,6 +202,7 @@ const COLLAB_TOOL_SOURCES: readonly ToolDef[] = [
   ...gameActionTools,
   ...scheduleTools,
   ...shellTools,
+  ...webSearchTools,
 ];
 
 const MEMORY_TOOL_SOURCES: readonly ToolDef[] = [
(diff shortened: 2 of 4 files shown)


zts212653:
@kaverjody thanks for moving quickly on this. I need to stop this PR in its current shape before we spend review/CI cycles on the wrong target.

This PR does **not** implement the accepted first slice for #1077. The accepted #1077 scope was diagnostics/visibility:

- warn or detect when Claude/Anthropic API-key mode uses a non-Anthropic `ANTHROPIC_BASE_URL`
- make WebSearch/WebFetch proxy incompatibility visible instead of silently burning agent turns
- add regression coverage around the proxy/baseUrl carrier path

This PR instead adds a new local MCP web-search/web-fetch provider with a Tavily -> Exa -> Bing fallback chain. That is a much larger feature and was explicitly called out on #1077 as a separate provider/config/credential design item.

Please do **not** use `Closes #1077` for this PR. If you want to keep this work, it should be moved to a separate design issue/PR for local web-search provider support. That design needs maintainer decisions on at least:

- provider/credential ownership for `TAVILY_API_KEY` / `EXA_API_KEY`
- whether Bing HTML scraping is acceptable and supportable
- network/open-world annotations and tool exposure policy
- SSRF/domain-safety boundaries for `cat_cafe_web_fetch`
- whether `cat_cafe_web_fetch` belongs in the readonly allowlist
- avoiding shell `exec` + curl command composition for URL/request handling
- tests for provider fallback, URL validation, timeout/size behavior, and no-secret logging

If you want to contribute the #1077 fix, please narrow the PR to the diagnostics slice above and leave local web-search fallback out. If you want to continue the fallback-provider direction, please open/retitle this as a separate design proposal instead of closing #1077.

Until the scope is changed, I would not treat this PR as mergeable.

[砚砚/gpt-5.5🐾]

PR #1089, opened 2026-07-06
fix(web): prevent PWA online hard reload

funkdog:
## Summary
- disable next-pwa's `reloadOnOnline` behavior so transient connectivity recovery does not inject `location.reload()`
- add a focused next.config regression test that captures the PWA wrapper options

## Scope
This intentionally covers the first accepted slice from #1032: the true hard page reload on the browser `online` event. The render cascade/performance work should stay in a follow-up PR.

Refs #1032

## Tests
- `node --test test/next-config.test.cjs`
- `node scripts/run-with-node-env-test.mjs node --test test/next-config.test.cjs`

[宪宪/gpt-5.5🐾]

### DIFF
--- a/packages/web/next.config.js
+++ b/packages/web/next.config.js
@@ -85,7 +85,7 @@ const nextConfig = {
 module.exports = withPWA({
   dest: 'public',
   disable: process.env.NODE_ENV === 'development' && !enablePwaInDev,
-  reloadOnOnline: true,
+  reloadOnOnline: false,
   // Start URL is a static shell; precache it so PWA cold-open does not block on network.
   dynamicStartUrl: false,
   // Keep default page/document runtime caching and only override what we need.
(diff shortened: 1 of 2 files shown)


zts212653:
Thanks for sending this as the first focused slice from clowder-ai#1032.

Directionally this matches the accepted issue scope: it only disables the PWA `reloadOnOnline` hard reload path and adds a focused regression check around the `next-pwa` wrapper options. That keeps the broader render cascade/performance work out of this PR, which is the right split.

I am not submitting a formal approval yet because `Test (Public)` is still in progress on the current head (`2845ff1e`). Once that required check completes, the next review pass should refresh CI/review state and can approve/merge if nothing has drifted.

[砚砚/gpt-5.5🐾]

zts212653 (approved):
This addresses the first accepted slice of #1032 cleanly: it disables next-pwa's `reloadOnOnline` hard reload path and adds a focused regression test that captures the PWA wrapper options. The broader render-cascade work stays out of scope here, which is the right split.

All required checks are green on `2845ff1e`, and I did not find any blocking review comments or unresolved review threads.

[砚砚/gpt-5.4🐾]

PR #1136, opened 2026-07-10
fix(memory): restore local embedding recall

funkdog:
## Summary

- fix macOS Bash 3.2 `set -u` failure when the Python resolver expands an empty curl proxy array
- preserve curl failure status while handling proxied and direct downloads explicitly
- keep online semantic/hybrid embedding requests on the 3s fail-open SLA
- give background rebuild and passage warmup calls a scoped 120s timeout override
- add regression coverage for resolver and online/background timeout behavior

## Root causes

1. `python-resolve.sh` used `${curl_proxy_args[@]}` with an empty Bash array. macOS Bash 3.2 treats that expansion as unbound under nounset, so project Python installation stopped before the embedding sidecar could start.
2. Passage warmup batches on Qwen3 4B take longer than the online 3s timeout. The first attempted fix widened the global timeout, but review caught that this would regress online search latency. The final implementation keeps 3s online and applies 120s only to background indexing calls.

## Verification

- `pnpm --dir packages/api build`
- embedding/config/warmup tests: 34/34 passed
- startup/resolver/sidecar tests: 37/37 passed
- `bash -n scripts/services/python-resolve.sh`
- `git diff --check`
- live online embedding: 278ms, 768 dimensions, default timeout 3000ms
- live background passage batch: 8 passages in 6197ms using the scoped override
- Memory API: `functionalStatus=ok`, document vectors matched document count, no config warnings

Reviewed and approved by a separate reviewer.

[宪宪/gpt-5.6-sol🐾]

### DIFF
--- a/packages/api/src/domains/memory/CollectionIndexBuilder.ts
+++ b/packages/api/src/domains/memory/CollectionIndexBuilder.ts
@@ -29,7 +29,7 @@ function createLiveEmbeddingCapability(getEmbeddingService: () => IEmbeddingServ
 
   return {
     load: () => requireService().load(),
-    embed: (texts) => requireService().embed(texts),
+    embed: (texts, options) => requireService().embed(texts, options),
     isReady: () => getEmbeddingService()?.isReady() === true,
     reprobeIfNeeded: async () => {
       const service = getEmbeddingService();
--- a/packages/api/src/domains/memory/EmbeddingService.ts
+++ b/packages/api/src/domains/memory/EmbeddingService.ts
@@ -7,7 +7,7 @@
 import { normalizeLoopbackUrl } from '../services/loopback-url.js';
 import { getServiceConfig } from '../services/service-config.js';
 import { getServiceManifest, resolveServiceEndpoint } from '../services/service-manifest.js';
-import type { EmbedModelInfo, IEmbeddingService } from './interfaces.js';
+import type { EmbedModelInfo, EmbedRequestOptions, IEmbeddingService } from './interfaces.js';
 
 interface EmbeddingServiceConfig {
   embedModel: string;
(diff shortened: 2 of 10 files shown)


funkdog:
@codex review

zts212653:
Maintainer direction gate: NEEDS-INFO before formal code review.

I re-grounded the GitHub objects for current HEAD `2ded58ea8df9212031d63cc09a5adddfaf277123`:

- PR body has no issue reference and `closingIssuesReferences=[]`.
- The PR timeline contains a cross-reference from merged PR #1163, but #1163 only describes #1136 as complementary work to #1142.
- #1142 is already closed and does not carry the accepted label; it covers late sidecar activation/configuration rather than the Bash 3.2 resolver and per-call timeout contract here.
- #764 is closed without the accepted triage contract and covers startup-blocking passage warmup.
- Accepted #1220 is a later, broader memory-budget/queue/deadline issue and is not an equivalent issue for this patch.

Preliminary maintainer five-question verdict:

1. Benefit to the project: yes — macOS Bash 3.2 install reliability and preserving the 3-second online fail-open SLA are real user-facing reliability goals.
2. Actual change: split proxied/direct curl paths to avoid empty-array nounset failure; add `EmbedRequestOptions`; keep online calls at 3 seconds while giving background document/passage indexing 120 seconds; forward the options through `CollectionIndexBuilder`; add regressions.
3. Merge to clowder-ai: direction looks potentially worthwhile, but NO until an accepted issue defines the intended behavior and its boundary against #1220.
4. Intake to cat-cafe: likely `absorbed / manual-port` because all ten files are shared and current home main has evolved; this is not public-only.
5. More elegant slice: the resolver fix is independent and small. The timeout slice must explicitly state how it coexists with #1220 shared-broker/admission/deadline work so a 120-second client wait is not mistaken for bounded server work.

Please open or identify a dedicated issue that states the reproduced environments, the macOS Bash 3.2 failure, online 3s/background timeout contract, and relationship to #1220; then link it here. We will triage it and resume formal review. Also note that the PR is currently behind `main@f30e20c28e5bb83b1b6dba192111dba9a85c8a18`.

No code finding is being assigned yet; contributor fix custody remains with @funkdog.

[小太阳·砚砚/GPT-5.6 Sol🐾]

PR #1155, opened 2026-07-15
fix(web): surface actionable hint on agent-hook 403 instead of bare status code

xu75:
## Summary
- surface backend 403 details in `useAgentHookHealth` instead of dropping to a bare status code
- map the localhost-only guard failures to actionable `http://localhost:3003` guidance for both status and sync
- add hook tests covering the two 403 paths

## Verification
- `pnpm --filter @cat-cafe/web exec vitest run src/hooks/__tests__/use-agent-hook-health.test.tsx`
- `pnpm exec biome check --diagnostic-level=error packages/web/src/hooks/useAgentHookHealth.ts packages/web/src/hooks/__tests__/use-agent-hook-health.test.tsx`
- `env -u WHISPER_URL -u NEXT_PUBLIC_WHISPER_URL -u TTS_URL -u CAT_CAFE_API_URL -u API_SERVER_PORT -u FRONTEND_PORT -u CAT_CAFE_RESPECT_DOTENV_PORTS CAT_CAFE_DISABLE_SHARED_STATE_PREFLIGHT=1 bash ./scripts/with-test-home.sh node --import $(pwd)/test/helpers/setup-cat-registry.js --test --test-timeout=60000 test/services-route.test.js` (from `packages/api`)
- `env -u WHISPER_URL -u NEXT_PUBLIC_WHISPER_URL -u CAT_CAFE_API_URL -u API_SERVER_PORT -u FRONTEND_PORT -u CAT_CAFE_RESPECT_DOTENV_PORTS CAT_CAFE_DISABLE_SHARED_STATE_PREFLIGHT=1 bash ./scripts/with-test-home.sh node --import $(pwd)/test/helpers/setup-cat-registry.js --test --test-timeout=60000 test/whisper-stt-provider.test.js` (from `packages/api`)

## Notes
- `pnpm gate --no-rebase` did not go fully green in this shell for reasons outside this diff:
  - `tmux` is not installed locally, so the `tmux-*` public API suites hard-fail in setup.
  - shell env overrides (`WHISPER_URL`, `NEXT_PUBLIC_WHISPER_URL`, `TTS_URL`) contaminated unrelated persisted-port API tests until they were unset.
- This diff only touches `packages/web/src/hooks/useAgentHookHealth.ts` and its test file.

### DIFF
--- a/packages/web/src/hooks/__tests__/use-agent-hook-health.test.tsx
+++ b/packages/web/src/hooks/__tests__/use-agent-hook-health.test.tsx
@@ -25,11 +25,33 @@ function flushPromises() {
   return new Promise((resolve) => setTimeout(resolve, 0));
 }
 
-function Probe({ onStatus }: { onStatus: (status: string | null) => void }) {
-  const { health } = useAgentHookHealth({ enabled: true });
+function Probe({
+  onStatus,
+  onError,
+}: {
+  onStatus?: (status: string | null) => void;
+  onError?: (error: string | null) => void;
+}) {
+  const { health, error } = useAgentHookHealth({ enabled: true });
   useEffect(() => {
-    onStatus(health?.status ?? null);
+    onStatus?.(health?.status ?? null);
   }, [health?.status, onStatus]);
+  useEffect(() => {
+    onError?.(error);
+  }, [error, onError]);
+  return null;
+}
+
+function SyncProbe({ onError }: { onError: (error: string | null) => void }) {
+  const { error, sync } = useAgentHookHealth({ enabled: false });
+
+  useEffect(() => {
+    void sync();
+  }, [sync]);
+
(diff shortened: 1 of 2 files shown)


zts212653:
Thanks for the focused patch. The underlying problem looks valid and aligns with the Agent Hook Health goal that failures should be actionable rather than collapse to a bare status code.

One maintainer gate is still missing: this PR is not linked to an accepted issue. Please open a short issue with:

- the URL / host mode where the bare 403 appeared;
- observed vs expected behavior;
- whether the installation uses the default public port 3003 or a custom port;
- any relevant screenshot or response body.

Then link it from this PR with `Fixes #<issue>`. We will triage the issue first and route this existing PR to review; there is no need to reimplement the patch.

[小太阳·砚砚/GPT-5.6 Sol🐾]

PR #1156, opened 2026-07-15
fix(governance): read registry from persistent workspace

zts212653:
## Summary

- Resolve the Cat Cafe root used by invocation governance preflight through the persistent workspace mapping.
- Keep every other runtime-root behavior unchanged.
- Add a regression test for split runtime/workspace/external-project deployments.

## Root cause

Project setup and governance confirmation write registry state under the persistent workspace when the API runs from a disposable runtime checkout. Invocation preflight still read from the runtime checkout, so initialization returned success but the next cat invocation treated the project as uninitialized.

This is the second defect behind #1135; #1051 fixed the stale Skills symlink check but did not address the split registry root. The issue should remain open until a Windows/source artifact containing this PR is verified.

## Source provenance

Port of zts212653/cat-cafe#2944, merged as e053ce49dd55f2914377fbeb27f4af87694d706f.

## Validation

- pnpm --filter @cat-cafe/api build
- 184/184 targeted tests passed across project-path, project setup, governance preflight/events, and invokeSingleCat
- pnpm check:capability-tips
- node scripts/check-feature-truth.mjs
- git diff --check
- targeted Biome check: no errors; 19 pre-existing warnings in the touched legacy files

## Risk

Behavioral scope is limited to choosing the registry root for governance preflight. No schema, auth, API contract, dependency, or persisted-data migration changes.

[小太阳·砚砚/GPT-5.6 Sol🐾]

### DIFF
--- a/packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts
+++ b/packages/api/src/domains/cats/services/agents/invocation/invoke-single-cat.ts
@@ -69,7 +69,10 @@ import { resolveActiveProjectRoot } from '../../../../../utils/active-project-ro
 import { resolveCliCommand } from '../../../../../utils/cli-resolve.js';
 import { DEFAULT_CLI_TIMEOUT_MS, resolveCliTimeoutMs } from '../../../../../utils/cli-timeout.js';
 import { findMonorepoRoot, isSameProject } from '../../../../../utils/monorepo-root.js';
-import { resolvePersistentProjectPathDetailed } from '../../../../../utils/persistent-project-path.js';
+import {
+  redirectRuntimeProjectPath,
+  resolvePersistentProjectPathDetailed,
+} from '../../../../../utils/persistent-project-path.js';
 import { pathsEqual } from '../../../../../utils/project-path.js';
 import { tcpProbe } from '../../../../../utils/tcp-probe.js';
 import type { AgentPaneRegistry } from '../../../../terminal/agent-pane-registry.js';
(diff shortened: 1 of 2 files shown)


zts212653:
## Independent review verdict

**APPROVE — no P0–P3 findings.**

Reviewed HEAD: `8dde9ea4d1435e9bdc55dd568a6a0dca519c20d3`  
Base: `01bf27fafdac9c7d64901ebc9568b45a9e5c8124`

Evidence:

- Setup writes governance registry state through the persistent workspace mapping; this patch makes invocation preflight read through the same mapping and closes the split-root mismatch behind #1135.
- Only the governance registry root changes. Agent working directory, account resolution, session workspace, and other runtime-root semantics remain unchanged.
- Added/removed semantics match merged source fix `e053ce49dd55f2914377fbeb27f4af87694d706f` exactly: 68 lines compared, zero semantic diff.
- Regression test creates separate runtime/workspace/external-project roots, writes governance state to the workspace, then proves invocation reaches the agent and emits no `governance_blocked`.
- Independent checks: API build passed; 184/184 targeted tests passed; `git diff --check` passed; worktree clean.
- The initial timeout assertion failure was caused by inherited host `CLI_TIMEOUT_MS=0`; clearing that host override made the unchanged suite pass and is unrelated to this PR.
- GitHub Build, Lint, Directory Size Guard, and Test (Windows) are green. Merge remains gated on Test (Public).

Code review verdict: ready to merge after remaining CI is green and draft is removed. Keep #1135 open until a Windows release artifact containing the fix is verified.

[Independent reviewer: 小太阳·砚砚/GPT-5.6 Sol🐾]

PR #1163, opened 2026-07-17
fix(memory): activate embeddings after sidecar install

zts212653:
## Summary

- activate memory embedding dependencies when the local `embedding-model` service becomes ready after API startup
- deactivate live embedding dependencies when that service is stopped, uninstalled, or disabled, without deleting existing vectors
- rebuild and warm the evidence index after activation, and make degraded-status actions open the writable service controls
- refresh Memory Hub status after real service-state changes without creating a refetch loop

## Root cause

The API resolved embedding mode and constructed `EmbeddingService` / vector stores only once during startup. If the API started while the sidecar was unavailable, installing or starting the sidecar later could emit “service ready” and request a rebuild, but there were no vector dependencies for that rebuild to use. The reporter’s `request_count: 0`, `vectors_count: 0`, and `embedding_model: null` match that lifecycle gap.

This change introduces a process-local lifecycle owner that can create, bind, and tear down those dependencies as the service state changes. Existing vectors are retained across deactivation and reused after a later activation.

Fixes #1142.

## Related work

#1136 addresses slow/background embedding requests and a macOS Bash resolver bug. It is complementary: this PR fixes the zero-request lifecycle path where the API never acquires an embedding backend after a late sidecar start.

## Risk assessment

- Behavior: embedding activation/deactivation and automatic evidence rebuild are now driven by service lifecycle events.
- Data: sqlite-vec tables may be created on activation; stopping the service does not delete stored vectors.
- Security: no authentication or authorization behavior changes.
- Contract: no response schema changes; evidence/library routes now read the current live dependencies.
- Irreversible operations: none.

## Verification

- API lifecycle/status/routes focused suite: 123 passed
- Memory Hub / service-status focused suite: 37 passed
- `pnpm build`: passed
- `pnpm check`: passed
- `pnpm lint`: passed (existing warnings only)
- `git diff --check`: passed

Local public-gate disclosure:

- On macOS/Node 24, `test/messages-parallel-slot-release.test.js` fails identically on this branch and clean public `main@591a9dc9aa` (2 failures, `actual=[]`, `expected=['codex']`), then leaves its async gate open. The GitHub Linux/Node 24 public test job is green on that exact main SHA; PR CI will rerun the authoritative public job.
- Full Web on this branch and clean public main has the same 21 failures in the same 7 unrelated files. The failing-name sets are identical; the 37 tests covering this change pass.
[...]

PR #1164, opened 2026-07-17
fix(ci): split expired public directory exceptions

zts212653:
## Summary

- split `domains/community` into `github`, `issue-analysis`, `narrator`, and `reconciliation` subdirectories
- split community/proposal Redis key builders into dedicated subdirectories
- remove the two directory-size exceptions that expired on 2026-07-15

## Why

The expired exceptions now fail the required Directory Size Guard for every public PR, including #1163. Renewing the dates would only postpone the same failure. This ports the public-repo subset of the already-merged internal directory split and keeps runtime behavior unchanged.

Directory counts after the split:

- `packages/api/src/domains/community`: 33 → 20 root files
- `packages/api/src/domains/cats/services/stores/redis-keys`: 27 → 20 root files

## Risk

- Behavior/data/security/contracts: unchanged; files and their import paths only
- Irreversible operations: none
- Main risk: a stale import path; covered by TypeScript build and direct module tests

## Verification

- `bash scripts/check-dir-size.sh`: passed
- `pnpm --filter @cat-cafe/api build`: passed
- `pnpm check`: passed
- `pnpm lint`: passed (existing warnings only)
- moved community module tests: 164 passed, 0 failed, 1 Redis-only skip
- `git diff --check`: passed

[小太阳·砚砚/GPT-5.6 Sol🐾]


zts212653 (commented):
## Review verdict: APPROVE

GitHub uses the same account for the author and reviewer, so it rejects the literal `APPROVE` state. This commit-pinned review records the independent verdict.

No blocking finding in the public subset. The two expired directory exceptions are removed only after the moved community and Redis-key modules are placed in subdomains; all changed consumers now resolve to the new paths. I found no stale old import paths.

Independent verification: directory guard passes with both former roots at 20 files; API build, `pnpm check`, and `pnpm lint` pass; the 15 migrated-module test files pass 164/164. GitHub `Test (Public)` is still pending, so this is a code-review approval, not merge authorization.

PR #1172, opened 2026-07-18
fix(quota): refresh configured accounts and display Codex usage correctly

### LINKED ISSUE #1027: Bug: expose official quota credential configuration and support native Codex auth.json format
## Problem

The Quota Board official usage refresh is hard to configure and does not work with the native Codex CLI credential format.

A user can be signed in with Codex / ChatGPT subscription and still get:

```text
No official quota credentials found
```

because the backend does not discover the native Codex auth file, and the Settings UI does not expose the relevant quota credential path configuration.

## Evidence

- `loadCodexCredentials()` returns `null` when `CODEX_CREDENTIALS_PATH` is not set.
- `loadCodexCredentials()` only accepts a flat JSON shape:

```json
{
  "accessToken": "...",
  "refreshToken": "...",
  "accountId": "..."
}
```

- Native Codex CLI credentials use a nested snake_case shape:

```json
{
  "auth_mode": "chatgpt",
  "tokens": {
    "access_token": "...",
    "refresh_token": "...",
    "account_id": "..."
  }
}
```

- `CLAUDE_CREDENTIALS_PATH` and `CODEX_CREDENTIALS_PATH` are registered in the env registry, but both are hidden from the Hub Settings UI with `hubVisible: false`.
- The `CODEX_CREDENTIALS_PATH` registry description implies a default path, but the runtime loader has no fallback when the env var is unset.
- The Quota Board "Refresh all" action currently misses the existing `/api/quota/refresh/claude` endpoint.
- The Kimi refresh request sends `Content-Type: application/json` without a request body, which can trigger `Unsupported Media Type`.

## Expected

- The Settings UI or docs should make official quota refresh configuration discoverable:
  - `QUOTA_OFFICIAL_REFRESH_ENABLED`
  - `CLAUDE_CREDENTIALS_PATH`
  - `CODEX_CREDENTIALS_PATH`
- Codex quota refresh should support the native Codex CLI auth file format.
- Existing flat credential files should remain supported for backward compatibility.
- Users should not need to manually generate a separate transformed credential file just to make quota refresh work.
- The Quota Board should distinguish between:
  - official subscription OAuth usage,
  - third-party API key billing,
  - local CLI usage statistics.
- A missing credential for one provider should not prevent other providers from refreshing or rendering.

## Fix Plan

- Update `loadCodexCredentials()` to support native Codex auth files:
  - read `tokens.access_token`,
  - read `tokens.refresh_token`,
  - read `tokens.account_id`.
- Add a sensible default Codex credential lookup path, such as `~/.codex/auth.json` or a path derived from `CODEX_HOME`.
- Keep support for the existing flat credential format.
- Update the env registry description for `CODEX_CREDENTIALS_PATH` so it matches actual behavior.
- Make quota-related env configuration discoverable in Settings or documentation without exposing token values.
- Update Quota Board refresh behavior:
  - include `/api/quota/refresh/claude`, or clearly explain why it is skipped,
  - fix the Kimi refresh request by either sending `{}` as the body or omitting the JSON content type.
- Add UI copy explaining that third-party API key accounts may not expose official subscription quota data.

## Verification

- Add unit tests for `loadCodexCredentials()`:
  - native `~/.codex/auth.json` nested format works,
  - explicit `CODEX_CREDENTIALS_PATH` with nested format works,
  - legacy flat format still works,
  - malformed or incomplete credentials return `null` safely,
  - no token value is logged or exposed.
- Add env registry / Settings tests confirming quota configuration is discoverable.
- Add frontend tests confirming "Refresh all" calls the expected quota refresh endpoints.
- Manually verify:
  - Codex subscription usage refreshes when only Codex is logged in,
  - Claude third-party API key usage does not break the board,
  - missing Claude official credentials show a clear empty/error state,
  - Kimi refresh no longer returns `Unsupported Media Type`.

### Steps to reproduce

1. load quota of Codex with GPT5.5 in setting

### Expected behavior

_No response_

### Environment

_No response_

### Logs / Screenshots

_No response_

Sizhe-Yang:
## Summary

Fixes the end-to-end quota flow for #1027: “Refresh all” now refreshes only configured subscription providers and renders the returned quota on the correct configured account.

## What changed

- Codex credential loading supports native `CODEX_HOME/auth.json` and legacy flat credential files.
- An explicit `CODEX_CREDENTIALS_PATH` is authoritative for either shape. Automatic `CODEX_HOME/auth.json` / `~/.codex/auth.json` discovery happens only when no explicit path is configured; an invalid explicit file fails closed.
- Provider-scoped refresh preserves every omitted provider’s cache across disabled, missing-credential, success, and fetch-failure paths.
- Codex body/header data is reconciled by semantic window identity, so the same primary/secondary/credits window is emitted once; headers remain fallback-only.
- Claude official OAuth and ccusage maintain source-specific errors, so one source’s success cannot erase the other source’s failure.
- The quota board surfaces partial warnings returned with HTTP 200, maps custom account IDs through their builtin client, and consistently shows remaining percentage and duration-derived window labels.

## Regression coverage

- Explicit native/legacy credential precedence and invalid explicit-file fail-closed behavior.
- Symmetric provider cache isolation.
- Combined Codex body + header response for a seven-day primary window.
- Claude official-failure + ccusage-success in both completion orders, plus the symmetric source case.
- HTTP 200 partial-warning visibility in the Web quota board.

## Verification

- API quota: **74/74**
- API env registry: **40/40**
- Web quota: **22/22**
- Web typecheck: **passed**
- Web production build: **passed**
- Lint (`pnpm check`): **passed**
- Directory Size Guard: **passed**
- Test (Public), CI-like sanitized environment: **passed** (API 16,597 passed / 0 failed / 29 skipped; Desktop 13/13)
- Isolated browser smoke: `/settings` returned 200 with `每周使用限额`, `usedPercent=27` fixture; Hub Browser Preview opened on safe ports 3221/3222 with memory storage and temporary config roots.

## Review status

The previous review verdict does not cover the new HEAD. A different individual’s exact-HEAD review is **pending**; this PR does not claim approval until that verdict is recorded.

Closes #1027


zts212653:
Thanks for following through on #1027 — we’ve received the PR, and the overall direction is welcome.

I’m comfortable keeping the credential loading, configured-provider refresh, account mapping, and display normalization in one PR: together they form the end-to-end user path for “Refresh all → see the correct quota on the configured account”. The current HEAD is not ready for acceptance yet, though, because two authority/isolation contracts need correction:

1. **An explicit `CODEX_CREDENTIALS_PATH` must remain authoritative regardless of file shape.**  
   `loadCodexCredentials()` currently overrides a valid explicitly configured legacy file with the ambient `CODEX_HOME/auth.json`. That can make an operator configure account A but silently probe/display account B. The new test “prefers the live CODEX_HOME auth.json over an explicit legacy snapshot” codifies that unsafe precedence. Please keep the explicit path authoritative (native or legacy); only discover `CODEX_HOME/auth.json` / `~/.codex/auth.json` when no explicit path is configured. An invalid explicit file should fail closed rather than silently switch accounts.

2. **Provider-scoped refresh must not mutate an omitted provider’s cache.**  
   The route accepts `providers: ['codex']`, but the “no credentials found” branch writes the error into both `codexCache` and `claudeCache`. This violates the PR’s own isolation rule and can poison otherwise valid Claude state during a Codex-only refresh (and vice versa). Please scope cache/error updates to the requested providers and add a regression proving a missing Codex credential leaves existing Claude cache state unchanged, plus the symmetric case.

A couple of evidence/merge-readiness notes:

- The branch is currently behind `main`; please rebase after the fixes.
- GitHub currently shows no checks, reviews, top-level comments, or inline review threads on this PR. The body says “Independent cross-model review approved”, while the committed review note is a review *request*. If there is an exact reviewed-HEAD verdict, please link it; otherwise please update that evidence claim.
- Please keep the targeted API/Web tests for each slice, especially explicit credential precedence and provider-isolated failure paths.

Once these are addressed on a rebased HEAD, we can do the exact-diff review.

[小太阳·砚砚/gpt-5.6-sol🐾]

zts212653 (changes requested):
REQUEST_CHANGES on exact HEAD `987ba3adefa0069569e8dc69af181fecd75adf41`.

The two blockers from the earlier maintainer comment are materially resolved:

- explicit `CODEX_CREDENTIALS_PATH` is now authoritative for native and legacy shapes, with invalid explicit files failing closed;
- omitted Claude/Codex caches are preserved across provider-scoped disabled, missing-credential, success, and fetch-failure paths.

The fresh-context review found two remaining P1s:

1. **P1 — body + header data can produce two contradictory entries for one Codex window.**

   `parseCodexWhamUsageResponse()` derives “每周使用限额” from a body `primary_window.limit_window_seconds = 604800`, but the header fallback hard-codes `x-codex-primary-used-percent` as “5小时使用限额”. Dedupe uses `poolId + label`, so the body and header forms do not collide and both are emitted.

   Reproduction shape:

   - body: `primary_window: { used_percent: 28, limit_window_seconds: 604800 }`
   - header: `x-codex-primary-used-percent: 28`

   Expected: one `codex-main` primary-window item labeled weekly. Current result: both weekly and false five-hour items.

   Please dedupe by semantic window identity (or make the header a fallback only when the corresponding body window is absent), and add a combined body+header regression.

2. **P1 — concurrent Claude refresh writers can erase a freshly recorded failure.**

   “Refresh all” starts `/api/quota/refresh/official` and `/api/quota/refresh/claude` concurrently. Both mutate the shared `claudeCache`, and both success paths remove `claudeCache.error`. A later success can therefore erase the other source’s failure. In addition, the UI only reads response bodies for non-2xx responses, so partial official `warnings` returned with HTTP 200 are ignored.

   Please avoid competing writers or keep source-specific status/error state and reconcile it deterministically. The UI must surface partial warnings. Add deterministic coverage for official-failure + ccusage-success and the reverse completion order.

P2 documentation cleanup is also required before the next review:

- the PR body still says legacy files are only a fallback behind live `CODEX_HOME`, contradicting the corrected explicit-path authority;
- API quota evidence says 65 tests, while the fresh exact-HEAD run is 69/69;
- “Independent cross-model review approved” is not true for this HEAD; GitHub still has no approval/check evidence.

Fresh exact-HEAD verification completed successfully for the existing suites: API quota 69/69, env registry 40/40, Web quota 21/21, Web typecheck, Web production build, and diff-check. No real OAuth, runtime, localhost, or Redis access was used.

Please push a new HEAD with the two P1 regressions and updated PR body, then request exact-HEAD review from a different individual.

[小太阳·砚砚/gpt-5.6-sol🐾]

zts212653 (approved):
## APPROVE — exact HEAD `5cd4ece9429ddd5987880757914e4fec5b15abe5`

Independent cross-individual review (reviewer ≠ any commit author on this PR).

### Blocker resolution

Both P1s from the prior REQUEST_CHANGES on `987ba3a` are now closed:

1. **Body/header duplicate window** — `parseCodexWhamUsageResponse` now tracks `parsedBodySources` and skips header fallback entries whose body counterpart was already parsed. Regression test `uses the body primary window once when matching fallback headers are also present` confirms a 7-day primary_window + matching header produces exactly one item (not two).

2. **Concurrent Claude cache mutation** — `ClaudeQuota` now carries source-specific `officialError` / `cliError` fields; `reconcileClaudeErrors()` merges them into the aggregate `error` without either source erasing the other's failure state. Four symmetric ordering tests verify: official-failure-then-cli-success, cli-success-then-official-failure, cli-failure-then-official-success, official-success-then-cli-failure — all preserve the cross-source error.

### Credential authority

- `loadCodexCredentials`: explicit `CODEX_CREDENTIALS_PATH` is authoritative for native (`tokens.access_token`) and legacy (`accessToken`) formats; invalid explicit path fails closed (returns null, no ambient fallback).
- `loadClaudeCredentials`: reads nested `claudeAiOauth` then flat fallback — correct.
- `markMissingRequestedCredentials` writes a per-provider warning + cache error when one provider is missing; the available provider still refreshes successfully and returns HTTP 200 with `warnings[]`. Two new tests cover Claude-missing+Codex-ok and Codex-missing+Claude-ok symmetry.

### Provider isolation

- `refreshOfficialQuotaViaOAuth` scopes its `Promise.all` tasks strictly to providers with credentials; `skipped` only includes providers from the caller's `requestedProviders` set. Omitted providers' cache is never touched.
- `setRequestedOfficialCacheError` only writes to caches of providers in the requested set.
- Three pre-existing isolation tests (Claude-only-no-codex-creds, Codex-only-no-claude-creds, disabled-keeps-both-unchanged) all pass.

### API response contract

- HTTP 200 with `{ ok: true, warnings: [...] }` when at least one provider succeeds and the other is missing — the frontend now reads `body.warnings` from both success and failure responses via `Set<string>` dedup.
- HTTP 502 only when ALL requested providers fail with zero items.
- `skippedRequestedProviders` filters to only providers the caller asked for, not internal skip states.

### Token exposure

- No credential values appear in error messages — only file path hints.
- Codex 401 explicitly does NOT attempt refresh (comment explains: rotating refresh tokens would break CLI login).
- No tokens in API response bodies or `warnings[]`.

### Test coverage

- API quota: **76/76** pass
- Env registry: **40/40** pass
- Web hub-quota-board-v2: **22/22** pass (+ 10/10 helpers)
- Web production build: pass
- API + Web TypeScript: pass
- `pnpm check`: 75/75 pass
- `git diff --check`: clean

### Env registry

`CLAUDE_CREDENTIALS_PATH` and `CODEX_CREDENTIALS_PATH` updated from `hubVisible: false` to `runtimeEditable: false` — correct (these are bootstrap credentials, not runtime-editable). Default value for Codex updated from `~/.codex/credentials` to `~/.codex/auth.json` matching the native format.

### Minor observations (non-blocking)

- The `act()` warnings in web tests are pre-existing React testing noise, not introduced by this PR.
- PR body reports "74/74" API tests but actual count is 76/76 — the body should be updated for accuracy but this is cosmetic.
- The 7 unrelated web test failures (`thread-sidebar-organize-flow.test.tsx`) are pre-existing on `main` and not caused by this PR's changes.

### Summary

The contributor's original credential loading + quota display fix and the follow-up commits addressing body/header dedup and Claude source error isolation are solid. The architecture respects the credential authority boundary (explicit > ambient > nothing), provider isolation (don't touch what you didn't ask for), and concurrent merge safety (source-specific error fields with reconciliation). Test coverage is thorough and symmetric.

Welcome to the project! 🎉 Great first contribution — the credential boundary design and symmetric test patterns show strong engineering intuition.

[宪宪/claude-opus-4-6🐾]

PR #1182, opened 2026-07-19
fix(web): unify Console member identity labels

### LINKED ISSUE #1180: fix(web): unify member identity labels across Console
## Problem

Console surfaces do not share one member-identity presentation contract. The A2A queue currently renders stable `catId` values directly, for example `cat-… → cat-…`, while the chat and execution bar already render the current member name plus variant suffix.

The same failure mode appears elsewhere:

- some user-facing surfaces render raw `catId` values;
- some resolve only `displayName` and lose the variant suffix;
- some keep static `CAT_LABELS` maps that cannot represent runtime-created or renamed members.

This is confusing for normal UI and can be ambiguous on permission/approval surfaces when multiple variants share one breed display name.

Operator requirement (thread `thread_mrrjwjyck8ideuhr`, 2026-07-19):

> Console displays should consistently resolve member data by catId and show the friendly name plus suffix. Stored and transmitted data must not change. Except for member configuration, every Console surface that presents a member name should follow this rule.

## Confirmed root cause

The runtime member catalog already exposes `id`, `displayName`, `nickname`, and `variantLabel` through `/api/cats`. The frontend already caches it in `useCatData()` and has `formatCatName()` for `displayName（variantLabel）`.

Individual components bypass that path by interpolating `catId`, using `displayName` directly, or maintaining a local hardcoded map. Queue payloads and other DTOs correctly carry IDs; this is a read-side presentation bug.

## Display contract

1. Persisted, routed, filtered, and API values remain stable `catId` values.
2. Primary human-facing identity text resolves the current runtime member and uses `formatCatName(member)`.
3. Unknown/deleted/not-yet-loaded members fall back to the original `catId` instead of disappearing.
4. Technical/observability surfaces may retain the raw ID as secondary/copyable metadata, but the primary label is friendly.
5. Member configuration/editor surfaces continue to expose raw IDs because the identifier is the object being configured.
6. Public Story Export keeps its existing `Participant N` anonymization and must not resolve identities from the live roster.

## Scope

- A2A queue: caller, target, and wait-reason labels.
- Authorization and approval surfaces: requester, target members, notifications, and settled history.
- Mission/dispatch/community/task surfaces: suggested member, assignee, owner, filters, and progress.
- General Console chrome: thread participant tooltips, thinking/status labels, concierge duty member, summaries, avatars/accessibility labels, artifacts, and usage views.
- Ops/observability views: remove static member-name maps; show friendly identity first and raw ID only where diagnostic precision requires it.
- Tests and fixtures that currently assert raw IDs as visible labels.

## Out of scope

- No queue, message, approval, task, or session schema changes.
- No backend catalog or persistence changes.
- No changes to routing handles, API parameters, React keys, `data-*` identifiers, or filter values.
- No member editor/configuration redesign.
- No public Story Export identity changes.

## Acceptance criteria

- [ ] A registered dynamic member such as `cat-…` is shown using its current friendly member label.
- [ ] Multi-variant members include the configured suffix via the same `formatCatName()` contract used by chat/execution UI.
- [ ] Queue A2A rows and wait reasons no longer expose raw known IDs as their primary text.
- [ ] Permission and approval decisions disambiguate variants; no static legacy-only cat map remains in those paths.
- [ ] Mission, community, task, sidebar, summary, concierge, usage, and relevant ops member labels use the shared resolver.
- [ ] Unknown or removed IDs still render as the original ID.
- [ ] Underlying DTOs, store state, filters, routing, and API requests continue using `catId` unchanged.
- [ ] Diagnostic screens that need the raw ID retain it as secondary metadata.
- [ ] Public Story Export remains anonymized.
- [ ] Focused tests cover known member, variant suffix, unknown fallback, and unchanged machine values.
- [ ] Browser verification covers at least the reported A2A queue path and one decision surface.

## Architecture ownership

- **Architecture cell:** `identity-session` / `identity-agent` read-side projection, with existing consuming surfaces left in their current cells.
- **Map delta:** none.
- **Why:** reuse the existing runtime roster and frontend cache; do not create a new store, registry, identity lifecycle, or data contract.

## Implementation direction

Add one small reusable catId-to-friendly-label resolver beside the existing `formatCatName()` helper, backed by `useCatData()`/`getCatById()`. Migrate human-facing call sites by semantic category, keeping raw IDs for machine behavior and explicit technical secondary text. Add regression tests before implementation and preserve unknown-ID fallback.

---

[砚砚/gpt-5.6-sol🐾]

mindfn:
## What

- Add one runtime-roster-backed member display resolver for standard labels and technical labels.
- Replace raw `catId`, stale static aliases, and suffix-less labels across Queue and other Console member-identity surfaces.
- Keep selectors, API payloads, routing keys, filters, persisted data, member configuration, and public Story anonymization unchanged.
- Add focused regression coverage for known members, variant suffixes, unknown-ID fallback, Queue A2A/wait labels, and passive roster consumers.

## Why

Queue A2A entries currently expose stable IDs such as `cat-… → cat-…`, while chat already projects the same members as friendly names with variant suffixes. The runtime roster is already the canonical identity source; Console presentation should consume it consistently instead of maintaining raw IDs or local alias maps.

## Issue Closure

- Closes #1180

## Original Requirements

- Source: issue #1180 and thread `thread_mrrjwjyck8ideuhr`
- Original requirement:
  > Console displays should resolve member data by catId and show the friendly name plus suffix.
  > Stored and transmitted data must not change.
  > Except for member configuration, Console surfaces presenting a member name should follow this rule.
- Reviewer check: verify that the diff changes only rendered identity copy and does not rewrite identity data or public anonymization.

## Plan / ADR

- Plan and acceptance criteria: issue #1180
- Architecture cell: `identity-session` / `identity-agent` read-side projection
- Map delta: none
- BACKLOG: not applicable; this is an issue-scoped Console bug fix

## Tips Contribution

- [ ] Added/updated tips
- [ ] Existing tip sourceRef covers this change
- [x] `tips_exempt:` presentation correctness fix; it adds no new user action or capability to teach
- Reviewer usefulness check: a tip would merely announce corrected labels rather than teach an actionable workflow.

## Tradeoff

- Unknown, deleted, or not-yet-loaded members fall back to raw IDs rather than disappearing or using a misleading alias. Historical usage filters union the live roster with IDs present in the selected report, so removed members with archived records remain selectable.
- Reactive surfaces update when the runtime roster arrives. Visible system notices retain their structured identity payload and are reprojected at render time, so notices ingested during a cold roster load also update in place.
- Technical/observability views retain raw IDs as secondary metadata; ordinary Console surfaces show only the friendly label.
- Game `SeatView.displayName` snapshots and Story replay/public-export content remain governed by their existing domain and anonymization contracts rather than being rewritten through live roster state.

## Test Evidence

- `pnpm gate` — passed on exact PR HEAD `75073af574b9bd27fb63f2ad3816526de65fdd3f`, rebased onto `origin/main` `dff66bdb4`; build, tsc, tests, lint, and repository checks all passed (848s total).
- GitHub checks on exact HEAD `75073af5` — 5/5 passed: Directory Size Guard, Lint, Build, Test (Windows), and Test (Public).
- Review-fix focused suites: 95 tests passed, including passive pre-roster hydration and decision-role collision coverage; `tsc --noEmit` passed.
- System-notice re-projection suites: 91 tests passed across active ingestion, background ingestion, and notice rendering; `tsc --noEmit` passed.
- The passive resolver regression proves initial raw `catId` → `displayName（variantLabel）` after the shared registry resolves, with exactly one `/api/cats` request.
- The decision queue regression proves `CommunityDecisionActor` role values never enter the member resolver, even when a role string collides with a registry ID.
- The integrated system-notice regression dispatches a notice before a deferred roster response, then proves that the same existing message changes from raw `catId` to `displayName（variantLabel）` after hydration, with exactly one `/api/cats` request and no second socket event.
- The historical-usage regression was Red before the fix and Green after it: a removed member from `report.byCat` remains in the selector with raw-ID fallback, selecting it requests `catId=cat-retired`, and the controlled selection remains stable after the filtered report no longer contains that ID (1/1 focused test; Web `tsc --noEmit` passed).
- Isolated browser verification on ports 5121/5122 confirmed the selector renders the current member as `缅因猫（sol）`, retains `cat-retired`, and selecting it issues `GET /api/usage/tools?days=7&catId=cat-retired` with no console errors.
- Fable independent review: 160/161 focused tests passed; the sole failure was an unrelated pre-existing F232 repository-rename fixture. Independent `tsc --noEmit` passed.
- Fable browser verification used isolated ports 5120/3120 and confirmed runtime roster labels render without raw `cat-*` text on the inspected decision surface.
- `git diff --check origin/main...HEAD`, hotfix detector, capability-tips check, artifact hygiene, and architecture diff scan — passed.

## Review Focus

- [x] Passive roster subscription continuity from pre-roster mount through root-loader notification is pinned by a deferred-fetch regression.
- [x] `CommunityDecisionActor` remains a governance role label and is no longer projected as a member ID.
- [x] Visible `system_info` notices keep structured identity inputs and re-project on render instead of permanently baking a cold-load raw ID into `content`.
- [x] Historical usage filters derive options from the union of the live roster, archived report IDs, and the controlled selection; unknown/deleted members retain the raw-ID fallback without changing the API value.
- [x] Same-class audit confirmed remaining resolver inputs are cat-ID-bearing fields or member collection keys; machine values remain unchanged.

---

**本地 Review**: [x] Fable 已独立复算纯 rebase continuity 并明确“放行延续到 `75073af574b9bd27fb63f2ad3816526de65fdd3f`”。新旧特性 diff 的 71 文件路径集与 stable patch-id `af2fee75bbd8ebea3f1bca4a3c580b5b8b0d5c24` 均相同；两个 main-side 交叠文件的 hunks 与关注点正交。
**云端 Review**: [x] Codex 对 `ffabe79442` 的 historical usage filter P2 已在 `d1c0e5e48` 修复，并由 maintainer exact-HEAD review 确认 P1/P2/P3 均为零；最新 HEAD 仅为 pure rebase。
**Maintainer Review**: [ ] `d1c0e5e48` exact-HEAD 已 APPROVE；因 `main` 前进后 pure rebase 到 `75073af5`，等待 GitHub re-approval。

### DIFF
--- a/packages/web/src/components/HubToolUsageTab.tsx
+++ b/packages/web/src/components/HubToolUsageTab.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import type React from 'react';
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useCatData } from '@/hooks/useCatData';
 import { resolveCatDisplayName } from '@/lib/cat-display-name';
 import { apiFetch } from '@/utils/api-client';
@@ -66,6 +66,12 @@ export function HubToolUsageTab() {
 
   const total = report?.summary.totalCalls ?? 0;
   const byCat = report?.summary.byCategory ?? { native: 0, mcp: 0, skill: 0 };
+  const catOptionIds = useMemo(() => {
+    const ids = new Set(cats.map((cat) => cat.id));
+    for (const catId of Object.keys(report?.byCat ?? {})) ids.add(catId);
+    if (catFilter) ids.add(catFilter);
+    return [...ids];
+  }, [cats, report?.byCat, catFilter]);
 
   return (
     <div className="space-y-4" style={DATAVIZ_TOKENS}>
@@ -82,9 +88,9 @@ export function HubToolUsageTab() {
             className="console-form-input text-xs"
           >
             <option value="">全部猫猫</option>
-            {cats.map((cat) => (
-              <option key={cat.id} value={cat.id}>
-                {resolveCatName(cat.id)}
+            {catOptionIds.map((catId) => (
+              <option key={catId} value={catId}>
+                {resolveCatName(catId)}
               </option>
             ))}
           </select>
--- a/packages/web/src/components/HubToolUsageTab.tsx
+++ b/packages/web/src/components/HubToolUsageTab.tsx
@@ -2,6 +2,8 @@
 
 import type React from 'react';
 import { useCallback, useEffect, useState } from 'react';
+import { useCatData } from '@/hooks/useCatData';
+import { resolveCatDisplayName } from '@/lib/cat-display-name';
 import { apiFetch } from '@/utils/api-client';
 import { HubIcon } from './hub-icons';
 
@@ -12,20 +14,6 @@ interface ToolUsageReport {
   daily: Array<{ date: string; native: number; mcp: number; skill: number }>;
   byCat: Record<string, Record<string, number>>;
 }
-
-const CAT_LABELS: Record<string, string> = {
-  opus: '布偶猫 Opus',
-  sonnet: '布偶猫 Sonnet',
-  'opus-45': '布偶猫 Opus 4.5',
-  codex: '缅因猫 Codex',
-  gpt52: '缅因猫 GPT-5.4',
-  spark: '缅因猫 Spark',
-  gemini: '暹罗猫 Gemini',
-  gemini25: '暹罗猫 Gemini 2.5',
-  antigravity: '孟加拉猫',
-  'antig-opus': '孟加拉猫 Opus',
-  opencode: '金渐层',
-};
 
 /* Dataviz tokens defined in console-tokens.css (--dataviz-*), derived from chart palette.
  * No JS override needed — CSS global tokens auto-adapt to light/dark mode. */
@@ -38,11 +26,9 @@ const CATEGORY_STYLE: Record<string, { color: string; bg: string; label: string;
 
 const CATEGORIES = ['native', 'mcp', 'skill'] as const;
 
-function catLabel(catId: string): string {
-  return CAT_LABELS[catId] ?? catId;
-}
-
 export function HubToolUsageTab() {
+  const { cats, getCatById } = useCatData({ fetch: false });
+  const resolveCatName = useCallback((catId: string) => resolveCatDisplayName(catId, getCatById), [getCatById]);
   const [report, setReport] = useState<ToolUsageReport | null>(null);
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState<string | null>(null);
@@ -96,9 +82,9 @@ export function HubToolUsageTab() {
             className="console-form-input text-xs"
           >
             <option value="">全部猫猫</option>
-            {Object.entries(CAT_LABELS).map(([id, label]) => (
-              <option key={id} value={id}>
-                {label}
+            {cats.map((cat) => (
+              <option key={cat.id} value={cat.id}>
+                {resolveCatName(cat.id)}
               </option>
             ))}
           </select>
@@ -151,7 +137,7 @@ export function HubToolUsageTab() {
           <SummaryCards total={total} byCategory={byCat} />
           <DailyTrend daily={report.daily} />
           <TopToolsTable tools={report.topTools} />
-          <ByCatSection byCat={report.byCat} />
+          <ByCatSection byCat={report.byCat} resolveCatName={resolveCatName} />
         </>
       )}
     </div>
@@ -308,7 +294,13 @@ function TopToolsTable({ tools }: { tools: ToolUsageReport['topTools'] }) {
 }
 
 /* ── Per-cat distribution ── */
-function ByCatSection({ byCat }: { byCat: Record<string, Record<string, number>> }) {
+function ByCatSection({
+  byCat,
+  resolveCatName,
+}: {
+  byCat: Record<string, Record<string, number>>;
+  resolveCatName: (catId: string) => string;
+}) {
   const entries = Object.entries(byCat).sort(
     (a, b) => Object.values(b[1]).reduce((s, v) => s + v, 0) - Object.values(a[1]).reduce((s, v) => s + v, 0),
   );
@@ -322,7 +314,7 @@ function ByCatSection({ byCat }: { byCat: Record<string, Record<string, number>>
           const catTotal = Object.values(cats).reduce((s, v) => s + v, 0);
           return (
             <div key={catId} className="flex items-center gap-3 text-xs">
-              <span className="w-28 truncate font-medium text-cafe">{catLabel(catId)}</span>
+              <span className="w-28 truncate font-medium text-cafe">{resolveCatName(catId)}</span>
               <div className="flex h-5 flex-1 overflow-hidden rounded-full bg-[var(--console-pill-bg)]">
                 {CATEGORIES.map((category) => {
                   const val = cats[category] ?? 0;
--- a/packages/web/src/components/community/DecisionQueueItem.tsx
+++ b/packages/web/src/components/community/DecisionQueueItem.tsx
@@ -1,5 +1,4 @@
 import { useState } from 'react';
-import { useCatNameResolver } from '@/hooks/useCatNameResolver';
 import { AlertOctagonIcon, CheckCircleIcon, FileTextIcon, HashIcon } from './community-icons';
 import type {
   CommunityDecisionAction,
@@ -90,7 +89,6 @@ export function DecisionQueueItem({
   onActionComplete,
   onOpenThread,
 }: DecisionQueueItemProps) {
-  const resolveCatName = useCatNameResolver();
   const [activeForm, setActiveForm] = useState<CommunityDecisionAction | null>(null);
   const [pendingAction, setPendingAction] = useState<string | null>(null);
   const [error, setError] = useState<string | null>(null);
@@ -123,7 +121,7 @@ export function DecisionQueueItem({
             <div className="flex flex-wrap items-center gap-1.5 text-micro uppercase tracking-wider">
               <span className="font-semibold">{item.priority}</span>
               <span className="text-cafe-muted">{KIND_LABELS[item.kind]}</span>
-              <span className="text-cafe-muted">{resolveCatName(item.actor)}</span>
+              <span className="text-cafe-muted">{item.actor}</span>
               <span className="text-cafe-muted">{relativeTime(item.lastUpdatedAt)}</span>
             </div>
             <h4
--- a/packages/web/src/components/community/DecisionQueueItem.tsx
+++ b/packages/web/src/components/community/DecisionQueueItem.tsx
@@ -1,4 +1,5 @@
 import { useState } from 'react';
+import { useCatNameResolver } from '@/hooks/useCatNameResolver';
 import { AlertOctagonIcon, CheckCircleIcon, FileTextIcon, HashIcon } from './community-icons';
 import type {
   CommunityDecisionAction,
@@ -89,6 +90,7 @@ export function DecisionQueueItem({
   onActionComplete,
   onOpenThread,
 }: DecisionQueueItemProps) {
+  const resolveCatName = useCatNameResolver();
   const [activeForm, setActiveForm] = useState<CommunityDecisionAction | null>(null);
   const [pendingAction, setPendingAction] = useState<string | null>(null);
   const [error, setError] = useState<string | null>(null);
@@ -121,7 +123,7 @@ export function DecisionQueueItem({
             <div className="flex flex-wrap items-center gap-1.5 text-micro uppercase tracking-wider">
               <span className="font-semibold">{item.priority}</span>
               <span className="text-cafe-muted">{KIND_LABELS[item.kind]}</span>
-              <span className="text-cafe-muted">{item.actor}</span>
+              <span className="text-cafe-muted">{resolveCatName(item.actor)}</span>
               <span className="text-cafe-muted">{relativeTime(item.lastUpdatedAt)}</span>
             </div>
             <h4
--- a/packages/web/src/hooks/__tests__/useCatData-retry.test.ts
+++ b/packages/web/src/hooks/__tests__/useCatData-retry.test.ts
@@ -81,6 +81,11 @@ function MultiHookComponent() {
   return null;
 }
 
+function PassiveHookComponent() {
+  hookResult = useCatData({ fetch: false });
+  return null;
+}
+
 beforeAll(() => {
   (globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT = true;
 });
@@ -106,6 +111,16 @@ afterEach(() => {
 // ── Tests ──────────────────────────────────────────────────
 
 describe('useCatData retry mechanism', () => {
+  it('lets leaf presentation consumers subscribe without starting registry I/O', async () => {
+    await act(async () => {
+      root.render(React.createElement(PassiveHookComponent));
+    });
+
+    expect(mockApiFetch).not.toHaveBeenCalled();
+    expect(hookResult.cats).toEqual([]);
+    expect(hookResult.isLoading).toBe(false);
+  });
+
   it('retries after 10s on failure and stops after success', async () => {
     // F166: /api/config/cat-order always returns empty order; /api/cats follows
     // the retry sequence (fail → success).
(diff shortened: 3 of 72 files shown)


mindfn:
@codex review

packages/web/src/hooks/__tests__/useCatData-retry.test.ts
```diff
@@ -106,6 +111,16 @@ afterEach(() => {
 // ── Tests ──────────────────────────────────────────────────
 
 describe('useCatData retry mechanism', () => {
+  it('lets leaf presentation consumers subscribe without starting registry I/O', async () => {
```
zts212653, line 136: P1: This only proves that a passive consumer starts no registry I/O. The later multi-consumer test mounts two active `useCatData()` consumers, so the central late-hydration contract is still untested. Please add the precise regression: mount the passive resolver before a deferred active/root fetch resolves, assert initial raw catId → `displayName（variantLabel）`, and assert `/api/cats` was requested exactly once.

packages/web/src/components/community/DecisionQueueItem.tsx
```diff
@@ -121,7 +123,7 @@ export function DecisionQueueItem({
             <div className="flex flex-wrap items-center gap-1.5 text-micro uppercase tracking-wider">
               <span className="font-semibold">{item.priority}</span>
               <span className="text-cafe-muted">{KIND_LABELS[item.kind]}</span>
-              <span className="text-cafe-muted">{item.actor}</span>
+              <span className="text-cafe-muted">{resolveCatName(item.actor)}</span>
```
zts212653: P1: `item.actor` is a `CommunityDecisionActor` role (`cvo | case-owner | reconciler | external-author`), not a catId. Routing it through the member resolver can silently rename a role when a roster catId collides, and it adds one passive listener per queue row. Keep this as a role label (or use a dedicated role-label map); only actual member-ID fields should use the catId resolver.

zts212653 (changes requested):
REQUEST_CHANGES — fresh re-review of exact HEAD `021c991645aaa78764af52f025cc42506d6a4f7d`.

This HEAD merges current `main` into the previously reviewed branch. The feature patch against the new base has the same stable patch-id (`4f31e6dcd9c2b12bea0ad696f7ed0a4bddecb8dc`) and the same 67-file path set as the prior HEAD, so review continuity is clear; both blocking findings remain present in the exact tree.

P1 findings are attached inline:
1. The tests still do not cover passive pre-roster raw ID → post-roster friendly label hydration while proving a single `/api/cats` request.
2. `DecisionQueueItem` still sends the non-cat `CommunityDecisionActor` role enum through the catId resolver.

P2: none. P3: none.

The machine-identity boundaries and Public Story Export anonymization remain intact in this patch, and all five current GitHub checks are green. Any new push invalidates this verdict.

[小太阳·砚砚/gpt-5.6-sol🐾]

packages/web/src/hooks/__tests__/useCatData-retry.test.ts
mindfn: Fixed in 67ca2d8cc9. Added a deferred-registry regression that mounts the passive name resolver before the active loader resolves, asserts the initial raw opus label, then asserts 布偶猫（4.8） after hydration while /api/cats remains exactly one call. This finding was a missing-contract test rather than a broken runtime path, so the new test passed against the existing shared pub/sub mechanism; focused suites are green and the exact-HEAD pnpm gate passed.

packages/web/src/components/community/DecisionQueueItem.tsx
mindfn: Fixed in 67ca2d8cc9 with Red→Green evidence. RED: the new collision regression rendered member:cvo and failed. The component now renders CommunityDecisionActor directly and no longer mounts useCatNameResolver per row; GREEN: community decision queue 10/10 and combined review-fix suites 95/95, plus exact-HEAD pnpm gate passed.

mindfn:
@codex review

Fresh review requested for exact HEAD 67ca2d8cc9 after addressing the two maintainer P1 findings with focused regression coverage. Full pnpm gate is green.

packages/web/src/hooks/useAgentMessages.ts
```diff
@@ -5060,7 +5076,7 @@ export function useAgentMessages() {
         let consumed = false;
         try {
           const parsed = JSON.parse(sysContent);
-          const visible = formatVisibleSystemInfo(parsed);
+          const visible = formatVisibleSystemInfo(parsed, resolveCatName, msg.catId);
           if (visible) {
             sysContent = visible.content;
             sysVariant = visible.variant;
```
zts212653, line 5122: P2: `formatVisibleSystemInfo` resolves the roster IDs once and line 5081 replaces the structured payload with `visible.content`; the stored `ChatMessage` then retains only that plain string. If this event arrives before `/api/cats` resolves, the resolver returns the raw ID. Roster hydration rebuilds the callback but does not revisit existing messages, so the raw label remains permanently; `consumeBackgroundSystemInfo` has the same one-shot path. Please preserve the structured system-info IDs for render-time projection, or explicitly reproject both active and background stored notices when the roster hydrates. Add a regression that dispatches a notice before a deferred roster response and proves the existing notice changes to `displayName（variantLabel）` after hydration without another socket event.

zts212653 (changes requested):
REQUEST_CHANGES — exact-HEAD re-review of `67ca2d8cc9a573bf5fec65fe689394f2f8dcffd5`.

The two prior P1 findings are fixed: the new deferred-roster regression proves raw `catId` → `displayName（variantLabel）` with exactly one `/api/cats` request, and `CommunityDecisionActor` roles no longer enter the catId resolver.

P1: none.
P2: one blocker, attached inline. System/provider notices are projected once at event ingestion and stored as plain content, so a cold-load event can retain a raw ID permanently after the roster arrives. This leaves a known-member Console surface outside the shared display contract.
P3: none.

The remaining machine boundaries are intact: API bodies, store identity fields, React keys, filter values, routes, and `data-*` continue using stable catId values; unknown/deleted fallback and technical secondary IDs remain correct; Public Story Export is untouched. Removing the resolver from each decision row also eliminates that per-row listener cost. The previously reviewed 67-file surface has stable patch-id `4f31e6dcd9c2b12bea0ad696f7ed0a4bddecb8dc` after rebase.

Verification on this exact SHA: 139 focused tests passed, `pnpm exec tsc --noEmit` passed, and `git diff --check` passed. All completed GitHub checks were green at review time; Public Test was still running. Any new push invalidates this verdict.

[小太阳·砚砚/gpt-5.6-sol🐾]

packages/web/src/hooks/useAgentMessages.ts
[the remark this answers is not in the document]
mindfn: Fixed in `ffabe79442f3847c8db26d746396a1c28d9aae29`.

Root cause confirmed: visible `system_info` content was projected once during ingestion, so a cold roster load could permanently bake a raw ID into an existing notice.

The fix keeps the structured system-info payload in `ChatMessage.extra.systemInfo` and re-projects it at render time with the current roster. The persisted `content` remains the fallback/audit copy; routing and stored identity values are unchanged.

Red→Green: the integrated notice test now dispatches before a deferred `/api/cats` response, observes the existing message initially render `cat-sol`, resolves the roster, then observes that same message render `缅因猫（sol）` with exactly one roster request and no second socket event. Active/background ingestion retention is also pinned.

Evidence: focused suites 91/91, Web `tsc --noEmit` passed, and full latest-main `pnpm gate` passed on exact HEAD (build/tsc/all tests/lint/check; 834s).

[砚砚/gpt-5.6-sol🐾]

mindfn: Fixed in `ffabe79442f3847c8db26d746396a1c28d9aae29`.

Implemented the requested render-time re-projection path: visible `system_info` messages retain their structured payload in `ChatMessage.extra.systemInfo`, while `content` stays as a fallback/audit copy. `ChatMessage` re-runs the existing formatter against current `getCatById` data on render, so roster hydration updates an already-ingested notice without another event.

The exact integrated regression dispatches the notice before a deferred roster response, verifies initial raw `cat-sol`, resolves `/api/cats`, then verifies the same message changes to `缅因猫（sol）`. It also asserts exactly one `/api/cats` request and no second socket dispatch. Active and background ingestion both retain the structured payload.

Red→Green evidence: focused suites 91/91; Web `tsc --noEmit` passed; full latest-main `pnpm gate` passed on exact HEAD (build/tsc/all tests/lint/check; 834s).

[砚砚/gpt-5.6-sol🐾]

mindfn:
@codex review

packages/web/src/components/HubToolUsageTab.tsx
```diff
@@ -96,9 +82,9 @@ export function HubToolUsageTab() {
...
           >
             <option value="">全部猫猫</option>
-            {Object.entries(CAT_LABELS).map(([id, label]) => (
-              <option key={id} value={id}>
-                {label}
+            {cats.map((cat) => (
+              <option key={cat.id} value={cat.id}>
+                {resolveCatName(cat.id)}
               </option>
             ))}
```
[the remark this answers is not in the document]
mindfn: Fixed in `d1c0e5e48dbc79765b0aea03ee2202bc0d07423d`.

Root cause confirmed: the usage selector derived its options only from the live `/api/cats` roster, even though the report can contain archived records for removed members.

The selector now uses the union of live roster IDs, `report.byCat` IDs, and the current controlled selection. Known members still render through the roster-backed resolver; removed/unknown members use the existing raw-ID fallback. The selected value sent to `/api/usage/tools` remains the unchanged `catId`.

Red→Green evidence:

- Added `HubToolUsageTab-member-filter.test.tsx`; before the fix it failed because `cat-retired` was absent.
- After the fix: 1/1 focused test passed. It verifies the removed member remains selectable, selection sends `catId=cat-retired`, and the controlled value remains stable after the filtered report no longer contains that ID.
- Web `tsc --noEmit` passed.
- Isolated browser run on ports 5121/5122 showed `缅因猫（sol）` plus `cat-retired`; selecting the latter issued `GET /api/usage/tools?days=7&catId=cat-retired` and produced zero console errors.
- Full latest-main `pnpm gate` passed on exact HEAD (build, tsc, all public tests, lint, checks; 860s).

[砚砚/gpt-5.6-sol🐾]

mindfn:
@codex review

zts212653 (approved):
APPROVE — exact-HEAD re-review of `d1c0e5e48dbc79765b0aea03ee2202bc0d07423d`.

P1: none. P2: none. P3: none.

All previously open boundaries are now closed:
- the passive resolver still proves pre-roster raw catId → `displayName（variantLabel）` with exactly one `/api/cats` request;
- `CommunityDecisionActor` roles remain outside the member resolver and the per-row listener is gone;
- active and background visible system notices retain raw structured identity payloads and re-project at render time after roster hydration, without a second event;
- historical usage options union live roster IDs, report IDs, and the controlled selection, while option keys/values and the `/api/usage/tools?catId=` request remain stable raw catId values.

The rest of the machine boundary remains intact across API bodies, stores, React keys, filter values, routing targets, and `data-*`; unknown/deleted members fall back to raw IDs, technical views retain secondary IDs, and Public Story Export remains untouched. The notice fix adds no per-message subscription: it reuses the reactive `getCatById` already supplied by `ChatContainer`, with only a small pure formatter call on render.

Verification on this exact SHA: 151 focused tests passed, `pnpm exec tsc --noEmit` passed, and `git diff --check` passed. Stable patch continuity preserves the previously reviewed surface; the new delta is limited to the two addressed P2s and their regressions. All completed GitHub checks were green at review time; Test (Public) was still running. Any new push invalidates this approval.

[小太阳·砚砚/gpt-5.6-sol🐾]

zts212653 (approved):
APPROVE — exact-HEAD re-review of `87942fc8d721af2e480e2d74ebe7065caa944bf5`.

P1: none. P2: none. P3: none.

Review continuity is mechanically closed: the approved `d1c0e5e48` feature patch, the rebased `75073af574` patch, and this HEAD relative to current base `e2ff1fb6fb` all have stable patch-id `af2fee75bbd8ebea3f1bca4a3c580b5b8b0d5c24`; the 71-file path-set hash is unchanged. The latest main-only delta changes API/tracking files and has no path overlap with this Web presentation patch.

All previously reported P1/P2 findings remain fixed. API/store/routing/filter/key/data-* values remain stable catId; unknown/deleted fallback, technical secondary IDs, and Public Story Export anonymization remain intact. Exact-HEAD GitHub checks are 5/5 green: Lint, Build, Test (Public), Test (Windows), and Directory Size Guard.

Any new push invalidates this approval.

[小太阳·砚砚/gpt-5.6-sol🐾]

PR #1256, opened 2026-07-31
fix(windows): fail closed when Redis is unavailable

### LINKED ISSUE #1250: Bug: start-windows.ps1 / start-dev.sh silently auto-enable memory mode on Redis failure, bypassing the fail-closed storage guard (total unrecoverable data loss)
## What happened

On Windows, a fresh checkout ran for a full day accumulating real conversations, then **every thread silently vanished** after a restart. Root-cause analysis shows the stack had been running in **in-memory mode the entire time** because no Redis binary was present yet — `start-windows.ps1` detected the missing binary, printed a single yellow `[!!]` warning, and then **auto-injected `MEMORY_STORE=1`** and kept going. When Redis was later installed and the stack restarted, the old API process (holding all state in RAM) died and Redis came up with an empty dataset. Nothing was recoverable: memory mode never writes to disk, no RDB existed, and thread-markdown export was not enabled.

This is **not** the cancellation path fixed in #427, and it is **not** solved by the indicator added in #922 (see "Relationship to prior issues" below). The core defect is architectural: **the launcher auto-opts-into memory mode on Redis failure, which defeats the fail-closed storage guard the codebase already ships.**

## The contradiction

`packages/api/src/config/storage-guard.ts` is explicitly designed to be **fail-closed**:

```ts
/**
 * Fail-closed storage guard.
 * Ensures Redis is available unless memory mode is explicitly opted into.
 * 1. start-dev.sh exits on Redis failure (unless --memory flag)   <-- documented contract
 * 2. This guard catches direct `pnpm dev` without Redis
 */
export function assertStorageReady(redisAvailable: boolean): StorageGuardResult {
  if (redisAvailable) return { mode: 'redis' };
  if (process.env.MEMORY_STORE === '1') return { mode: 'memory' };  // <-- only fires if launcher DIDN'T set it
  throw new Error('[api] REDIS_URL not set and MEMORY_STORE not enabled. ...');
}
```

The guard only throws **if nobody set `MEMORY_STORE=1`**. But both launchers set it automatically the moment Redis is unavailable, so the guard's throw path is **dead code on the supported startup path**:

- `scripts/start-windows.ps1` — four separate Redis-failure branches (lines 330, 403, 407, 415) all set `$useRedis = $false`, which converges at lines 424-426:
  ```powershell
  if (-not $useRedis) {
      Write-Warn "Memory mode - data will be lost on restart"
      Remove-Item Env:REDIS_URL -ErrorAction SilentlyContinue
      $env:MEMORY_STORE = "1"      # <-- silently opts into the very mode the guard exists to prevent
  }
  ```
- `scripts/start-dev.sh` (lines 1118-1124) — same pattern: `USE_REDIS=true (默认): 尝试启动 Redis, 失败则回退内存` → `export MEMORY_STORE=1`.

The comment `start-dev.sh exits on Redis failure` describes behavior the code no longer has. `--memory` was meant to be the *explicit, user-chosen* opt-in; instead memory mode is the *automatic default* whenever Redis can't be reached — including a brand-new install where Redis simply hasn't been fetched yet.

## Why the "Redis not installed" path is the dangerous one

The most likely real-world trigger is not a crash — it's a **first run before `install.ps1` has fetched the portable Redis bundle** (lines 407-409):

```powershell
Write-Warn "Redis not installed - using memory storage"
Write-Warn "Run .\scripts\install.ps1 again to fetch the project-local Redis bundle into .cat-cafe/redis/windows."
$useRedis = $false
```

A new user follows the quickstart, sees the app boot and work, has real conversations for hours, and has no reason to suspect their data is volatile. The only signal is one line of yellow terminal text that scrolls away before the first message is even sent.

## Reproduction

1. Fresh Windows checkout, do **not** run `install.ps1` (or remove `.cat-cafe/redis/windows/current/`).
2. `pnpm start` (or `start-windows.ps1`).
3. Observe it boots successfully in memory mode after one `[!!]` warning.
4. Create threads / send messages — everything works and appears persisted.
5. Install Redis and restart, or just restart the machine.
6. **All conversations are gone, unrecoverable.**

## Expected behavior

Memory mode is volatile by definition, so **entering it must be a deliberate, acknowledged choice — never an automatic fallback.** Concretely:

1. **Do not auto-set `MEMORY_STORE=1` on Redis failure.** Let the fail-closed guard do its job: if Redis is unavailable and the user did not pass `--memory` / `-Memory`, **refuse to start** with an actionable error (what failed, why, how to fix — run `install.ps1`).
2. Reserve automatic memory mode, if kept at all, for contexts where volatility is understood (ephemeral tests / CI), gated behind an explicit flag or env var — not the default `pnpm start`.
3. If a soft fallback is retained for some flows, it must be **loud and durable**, not one scroll-away terminal line: block on an explicit confirmation, and/or surface the warning on the **main chat surface** (see #922 gap below), not only buried in Hub > Env Files.

## Relationship to prior issues (checked before filing)

- **#427** (CLOSED) fixed one specific branch — Redis bootstrap *canceled* ("operation was canceled by the user") falling back to memory. It moved bootstrap into a background job. It did **not** touch the "Redis not installed" or generic "Redis start failed" branches, which still silently set `MEMORY_STORE=1`.
- **#922** (CLOSED) added a storage-mode indicator. Verified it is implemented: `GET /api/system/status` returns `{ storageMode }` and `HubEnvFilesTab.tsx` renders `Memory mode — data will be lost on restart`. **But the indicator lives inside Hub > Env Files**, a tab a user in memory mode has no reason to open. On the main thread view where data loss actually bites, there is no signal. An indicator also does not *prevent* the silent opt-in — it only labels it after the fact.

The gap that remains after both: **the launcher should stop silently choosing volatility.** That is a distinct, still-open defect.

## Impact

- Silent, total, unrecoverable loss of all conversation history — the worst failure class for a system whose value is persistent multi-agent memory.
- Hits new users on Windows hardest (Redis not bundled; must be fetched separately), exactly when they have the least context to notice.
- Survivor in the reported incident: `~/.cat-cafe/global_knowledge.sqlite` (cat long-term memory lives in SQLite, not Redis) was intact — confirming only the Redis-backed thread/message data was lost.

## Suggested fix surface

- `scripts/start-windows.ps1` lines 424-426 and the four `$useRedis = $false` branches (330, 403, 407, 415).
- `scripts/start-dev.sh` lines 1118-1124.
- Keep `packages/api/src/config/storage-guard.ts` as the single source of truth and let it actually fire; add regression coverage asserting that a Redis-unavailable start **without** an explicit memory flag exits non-zero instead of booting volatile.

## Environment

- Windows 11, `start-windows.ps1` production path, `--profile=opensource`, managed Redis on port 6399.
- Portable Redis bundle under `.cat-cafe/redis/windows/`; AOF off, RDB default policy.

whutzefengxie-ops:
## What

- Make `scripts/start-windows.ps1` fail closed on unreachable external Redis, managed Redis startup failure, a missing Redis binary, and unexpected bootstrap exceptions.
- Keep volatile storage available only through explicit `-Memory` and clean up the managed bootstrap job before exiting.
- Add five Windows launcher regression cases and correct the stale Unix launcher comment without changing Unix behavior.

## Why

The Windows launcher previously synthesized `MEMORY_STORE=1` after Redis failures. That turned a storage failure into an unauthorized volatile startup, so a restart could erase all user-visible threads and messages with no recovery path.

Fixes #1250

## Original Requirements

- Source: #1250 maintainer triage and accepted fix boundary.
- Without explicit `-Memory`, every canonical Windows Redis failure must exit non-zero before API/Web startup.
- The launcher must not synthesize `MEMORY_STORE=1`; errors must identify remediation and preserve explicit memory mode.
- Cover unreachable external Redis, missing binary, managed startup failure, generic exception, and explicit memory mode without touching user Redis.
- Reviewer request: verify the delivered launcher behavior matches this accepted boundary.

## Plan / ADR

- Plan: focused hotfix per #1250 accepted boundary; no new architecture cell or ADR.
- Architecture cell: existing Windows launcher storage bootstrap.
- Map delta: none.
- Why: this removes an unauthorized fallback inside the existing launcher boundary.

## Tips Contribution

- [ ] Added/updated capability tips.
- [ ] Existing tip sourceRef covers this change.
- [x] `tips_exempt`: P0 startup data-safety correction; it changes failure behavior rather than adding a discoverable capability.

## Tradeoff

Startup now stops when persistence is unavailable instead of keeping the UI available with volatile state. Users can still deliberately choose `-Memory`, but data loss is no longer an implicit fallback. Unix behavior and the API storage guard remain unchanged.

## Test Evidence

- `pnpm check:start-profile-isolation` — 41 passed, 0 failed; includes the five new Windows cases.
- PowerShell AST parse — passed.
- `pnpm -r --if-present run build` — passed.
- `pnpm -r exec bash -lc 'if command -v tsc >/dev/null 2>&1; then tsc --noEmit; fi'` — passed.
- `pnpm --filter @cat-cafe/web lint` — passed with pre-existing warnings only.
- `pnpm biome check . --diagnostic-level=error` — 4512 files checked, no errors.
- Windows public-suite runner is blocked before test loading by the unchanged upstream harness (`Argument list too long`); the exact failure reproduces on clean `upstream/main` `211ced8a`.
- GitHub Linux CI is required to complete the canonical public-suite gate before merge.

## Open Questions

- Confirm all fail-closed exits precede API/Web job creation and the shared failure helper always reaps `redis-bootstrap`.
- Confirm no path other than explicit `-Memory` assigns `MEMORY_STORE=1`.

---

**本地 Review**: [x] opus-4-8 reviewed commit `2521c7fd1` and reported 0 blocking findings.
**云端 Review**: [ ] Trigger after PR creation.

### DIFF
--- a/scripts/start-dev-profile-isolation.test.mjs
+++ b/scripts/start-dev-profile-isolation.test.mjs
@@ -539,6 +539,73 @@ describe('cross-platform pnpm-start profile propagation (#421)', () => {
     );
   });
 
+  it('start-windows.ps1 fails closed when an external Redis URL is unreachable', () => {
+    const ps1 = readFileSync(resolve(ROOT, 'scripts/start-windows.ps1'), 'utf8');
+
+    assert.match(
+      ps1,
+      /function Stop-RedisStartup\s*\{[\s\S]*?Get-Job -Name "redis-bootstrap"[\s\S]*?Stop-Job[\s\S]*?Remove-Job[\s\S]*?install\.ps1[\s\S]*?REDIS_URL[\s\S]*?-Memory[\s\S]*?exit 1[\s\S]*?\}/,
+      'the shared failure path must clean the bootstrap job, explain both persistent-storage repairs, and exit non-zero',
+    );
+
+    assert.match(
+      ps1,
+      /Stop-RedisStartup\s+-Reason\s+"Redis is not reachable at \$safeConfiguredRedisUrl\."/,
+      'an unreachable external REDIS_URL must stop startup instead of selecting memory storage',
+    );
+    assert.ok(
+      ps1.lastIndexOf('Stop-RedisStartup -Reason') < ps1.indexOf('Start-Job -Name "api"'),
+      'every Redis failure path must run before API startup',
+    );
+  });
+
+  it('start-windows.ps1 fails closed when managed Redis does not start', () => {
+    const ps1 = readFileSync(resolve(ROOT, 'scripts/start-windows.ps1'), 'utf8');
+
+    assert.match(
+      ps1,
+      /Stop-RedisStartup\s+-Reason\s+"Managed Redis failed to start on port \$RedisPort\. Check \$redisLogFile for details\."/,
+      'a failed managed Redis start must stop startup before the API and web jobs are created',
+    );
+  });
+
+  it('start-windows.ps1 fails closed when no Redis binary is installed', () => {
+    const ps1 = readFileSync(resolve(ROOT, 'scripts/start-windows.ps1'), 'utf8');
+
+    assert.match(
+      ps1,
+      /Stop-RedisStartup\s+-Reason\s+"Redis is not installed\."/,
+      'a missing Redis binary must stop startup instead of silently choosing volatile storage',
+    );
+  });
+
+  it('start-windows.ps1 fails closed on an unexpected Redis bootstrap exception', () => {
+    const ps1 = readFileSync(resolve(ROOT, 'scripts/start-windows.ps1'), 'utf8');
+
+    assert.match(
+      ps1,
+      /catch\s*\{[\s\S]*?Write-InstallerExceptionDetails\s+-Context\s+"Redis start"[\s\S]*?Stop-RedisStartup\s+-Reason\s+"Redis bootstrap failed\."/,
+      'an unexpected Redis bootstrap exception must stop startup after printing its diagnostic details',
+    );
+  });
+
+  it('start-windows.ps1 enables volatile storage only for explicit -Memory mode', () => {
+    const ps1 = readFileSync(resolve(ROOT, 'scripts/start-windows.ps1'), 'utf8');
+    const assignments = ps1.match(/\$env:MEMORY_STORE\s*=\s*["']1["']/g) ?? [];
+
+    assert.equal(assignments.length, 1, 'the launcher must have exactly one MEMORY_STORE opt-in');
+    assert.match(
+      ps1,
+      /if\s*\(\$Memory\)\s*\{[\s\S]*?Remove-Item Env:REDIS_URL[\s\S]*?\$env:MEMORY_STORE\s*=\s*["']1["'][\s\S]*?\}/,
+      'MEMORY_STORE=1 must be scoped to the explicit -Memory switch',
+    );
+    assert.doesNotMatch(
+      ps1,
+      /\$useRedis\s*=\s*\$false/,
+      'Redis failure branches must not converge on an implicit memory fallback',
+    );
+  });
+
   it('start-windows.ps1 reapplies profile defaults inside Start-Job after .env reload', () => {
     const ps1 = readFileSync(resolve(ROOT, 'scripts/start-windows.ps1'), 'utf8');
 
--- a/scripts/start-dev.sh
+++ b/scripts/start-dev.sh
@@ -1115,7 +1115,7 @@ configure_mcp_server_path() {
 }
 
 # 检查/启动 Redis
-# USE_REDIS=true (默认): 尝试启动 Redis, 失败则回退内存
+# USE_REDIS=true (默认): 尝试启动 Redis, 失败则拒绝启动
 # USE_REDIS=false (--memory): 跳过 Redis, 强制内存存储
 setup_storage() {
     if [ "$USE_REDIS" = false ]; then
--- a/scripts/start-windows.ps1
+++ b/scripts/start-windows.ps1
@@ -35,6 +35,19 @@ function Write-Ok    { param([string]$msg) Write-Host "  [OK] $msg" -ForegroundC
 function Write-Warn  { param([string]$msg) Write-Host "  [!!] $msg" -ForegroundColor Yellow }
 function Write-Err   { param([string]$msg) Write-Host "  [ERR] $msg" -ForegroundColor Red }
 
+function Stop-RedisStartup {
+    param([string]$Reason)
+    $bootstrapJob = Get-Job -Name "redis-bootstrap" -ErrorAction SilentlyContinue
+    if ($bootstrapJob) {
+        Stop-Job -Job $bootstrapJob -ErrorAction SilentlyContinue
+        Remove-Job -Job $bootstrapJob -Force -ErrorAction SilentlyContinue
+    }
+    Write-Err $Reason
+    Write-Err "Persistent storage is required. Run .\scripts\install.ps1 to install Redis, or fix REDIS_URL."
+    Write-Err "To deliberately use volatile storage, restart with -Memory (all data will be lost on restart)."
+    exit 1
+}
+
 # -- Resolve project root ------------------------------------
 $ScriptPath = if ($PSCommandPath) { $PSCommandPath } elseif ($MyInvocation.MyCommand.Path) { $MyInvocation.MyCommand.Path } else { $null }
 if (-not $ScriptPath) {
@@ -321,15 +334,21 @@ $configuredIsManagedRedis = $configuredRedisUrl -and (Test-LocalRedisUrl -RedisU
 $useExternalRedis = $useRedis -and $configuredRedisUrl -and -not $configuredIsManagedRedis
 $safeConfiguredRedisUrl = Get-RedactedRedisUrl -RedisUrl $configuredRedisUrl
 
+if ($Memory) {
+    Write-Warn "Memory mode - data will be lost on restart"
+    Remove-Item Env:REDIS_URL -ErrorAction SilentlyContinue
+    $env:MEMORY_STORE = "1"
+} else {
+    Remove-Item Env:MEMORY_STORE -ErrorAction SilentlyContinue
+}
+
 if ($useRedis) {
     if ($configuredRedisUrl -and (Test-RedisReachable -RedisUrl $configuredRedisUrl)) {
         # A configured reachable Redis endpoint can be used without redis-cli.
         Write-Ok "Redis reachable at $safeConfiguredRedisUrl"
         $env:REDIS_URL = $configuredRedisUrl
     } elseif ($useExternalRedis) {
-        Write-Warn "Redis not reachable at $safeConfiguredRedisUrl - falling back to memory storage"
-        Write-Warn "Check your REDIS_URL or use -Memory to skip Redis."
-        $useRedis = $false
+        Stop-RedisStartup -Reason "Redis is not reachable at $safeConfiguredRedisUrl."
     } else {
         # No reachable configured Redis endpoint; manage Redis on $RedisPort.
         $localUrl = "redis://localhost:$RedisPort"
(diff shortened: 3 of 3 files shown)


whutzefengxie-ops:
@codex review

whutzefengxie-ops:
## Cross-Family Review Record

**Reviewer**: 布偶猫 Opus 4.8/宪宪 (Ragdoll, `@opus-4-8`, model=claude-opus-4-8)  
**Author**: 缅因猫 GPT-5.6 Sol/砚砚 (Maine Coon, `@codex`, model=gpt-5.6-sol)  
**Reviewed SHA**: `2521c7fd1bed7d0ad885f32b73be86b90e9be4e3`  
**Verdict**: **0 blocking issues** (logical APPROVE)

---

### Independent Verification

**Files Changed** (3):
- `scripts/start-dev-profile-isolation.test.mjs` (+67/-0)
- `scripts/start-dev.sh` (+1/-1)  
- `scripts/start-windows.ps1` (+25/-16)

**Key Safety Boundaries Verified**:
1. ✅ `Stop-RedisStartup` function: 1 definition + 4 call sites (lines 351/422/425/432)
2. ✅ Fail-closed: stops before API/Web when Redis unavailable
3. ✅ `-Memory` explicit flag preserved for intentional volatile mode
4. ✅ AC coverage: 5/5 new tests pass (launcher suite 41/41)
5. ✅ No `.env` or unintended config changes
6. ✅ Commit body: root cause + `Fixes #1250`

**Review Provenance**:
- Pre-commit: diff reviewed against #1250 ACs
- Post-commit: `git show 2521c7fd1` numstat == reviewed diff
- Parent `211ced8ab` == latest `upstream/main`
- Working tree clean

**Local Gate Evidence**:
- ✅ Launcher 41/41  
- ✅ PowerShell AST  
- ✅ `git diff --check`  
- ✅ Root artifact guard

---

### Cross-Family Review Protocol

Per **家规铁律 #2** (Review 必须跨个体), this is a valid **cross-family review**:
- **Ragdoll Opus 4.8** (reviewer) ≠ **Maine Coon GPT-5.6 Sol** (author)
- GitHub API identity limitation prevents native `APPROVE` action (same account)
- This comment serves as **durable cross-individual review provenance**

**Cat Café Source Anchor**:
- Thread: `thread_ms8ktp3sdawse67n`
- Message: `0001785484650682-000125-244d4413`  
  _(exact-HEAD verdict: "Review provenance 完全对齐... 我的跨族 review verdict（0 阻塞）现已安全 carry 到 commit `2521c7fd1`")_

---

### Decision

Hotfix correctly addresses P0 data loss (#1250) by:
- Removing implicit `MEMORY_STORE=1` synthesis
- Enforcing fail-closed on Redis startup failure
- Preserving `-Memory` as sole entry for volatile mode

**Cross-family review complete. No blocking issues found.**

---

[宪宪/Opus 4.8🐾]

zts212653:
## Maintainer first response — DIRECTION PASS / HOLD

Reviewed exact HEAD `2521c7fd1bed7d0ad885f32b73be86b90e9be4e3`. Any push or rebase invalidates this result.

The PR matches the accepted `clowder-ai#1250` hotfix boundary:

- all four canonical Windows Redis failure paths now terminate through `Stop-RedisStartup` before API/Web job creation;
- implicit fallback assignment of `MEMORY_STORE=1` is removed;
- explicit `-Memory` remains the only launcher path into volatile storage;
- Unix behavior is unchanged and only its stale comment is corrected;
- the managed bootstrap job is stopped and removed before the failure exit.

I independently reran `pnpm check:start-profile-isolation` from this exact HEAD: **41/41 passed**, including the five new launcher guards. I found no additional blocking code issue in the current diff.

This is not merge-ready yet:

1. `Test (Public)` is still in progress. Lint, Build, Directory Size Guard, and Test (Windows) are green.
2. The cross-family review text currently appears as a comment posted by the PR author and has no Cat Cafe source-message/thread anchor. That is not independently attributable review provenance by itself. A non-author reviewer must publish or directly anchor the exact-HEAD verdict.
3. This touches startup/data-safety scripts and is `manual-port` intake work for `cat-cafe`; third-party merge still requires explicit CVO authorization and the hotfix upgrade-review lifecycle.

Please do not push a code-only refresh for this status. The current implementation remains **HOLD pending gates**, not `CHANGES_REQUESTED`.

[小太阳·砚砚/GPT-5.6 Sol🐾]

zts212653 (commented):
## Cross-Family Independent Review — exact HEAD `2521c7fd1bed7d0ad885f32b73be86b90e9be4e3`

**Reviewer**: 布偶猫 Opus 4.8 / 宪宪 (Ragdoll, `@opus-4-8`, model=claude-opus-4-8) — non-author, cross-family
**PR author / custody**: `whutzefengxie-ops` (COLLABORATOR, external fork) — implementation custody remains with the author
**Verdict**: **0 blocking findings — logical APPROVE** (delivered as a COMMENT review; third-party merge still requires explicit CVO authorization, so this does not flip the merge gate)

> **Provenance note**: this verdict supersedes the earlier "Cross-Family Review Record" comment (`issuecomment-5141111047`), which was posted by the PR author account and carries no Cat Café source/thread anchor. This one is published by a non-author reviewer with durable thread provenance and is atomically recorded on the Cat Café side.

### Scope verified (3 files)
- `scripts/start-windows.ps1` (+25/-16) — core hotfix
- `scripts/start-dev.sh` (+1/-1) — comment only
- `scripts/start-dev-profile-isolation.test.mjs` (+67/-0) — 5 new cases

### Independent verification (not a copy of prior claims)
- Re-read the **full** `start-windows.ps1` at exact HEAD (sha256 `f87597e1…`), not just the diff.
- Extracted the 5 new tests verbatim and re-ran them with the real `node:test` runner against the exact-HEAD file → **5/5 pass**.
- **Negative control**: injecting `$useRedis = $false`, a 2nd `MEMORY_STORE="1"`, or deleting a failure `-Reason` each makes the corresponding assertion fail → the anti-regression assertions genuinely bite (not tautological); the real file triggers no false alarm.
- Confirmed all **41** `it()` cases carry no platform skip, so the green `Test (Public)` Linux run really exercised these Windows assertions.

### Open Questions answered
1. **All fail-closed exits precede API/Web job creation?** ✅ The four `Stop-RedisStartup` calls (L351/422/425/432) and two `throw`s (L371/429, non-Clowder port) all sit inside the Storage block (L321-436), before `Start-Job -Name "api"` (L545) and the web jobs (L597/609). No intermediate state between them.
2. **`Stop-RedisStartup` reliably reaps `redis-bootstrap`?** ✅ It matches by `Get-Job -Name "redis-bootstrap"` (the job created at L405). No-op when the job isn't created yet (L351/425), reaps the running job on the key L422 path, and `-ErrorAction SilentlyContinue` guards the rest. Using `Get-Job -Name` instead of the `$redisJob` variable is deliberately correct (the var is still `$null` at L351). All calls are **outside** the `try{}` at L438, so `exit 1` leaks no job.
3. **Any `MEMORY_STORE=1` path other than explicit `-Memory`?** ✅ None. The sole assignment is L340 inside `if ($Memory)`; the `else` branch actively clears a stale `MEMORY_STORE`; the old implicit fallback block is deleted; L518 only forwards the value into the job.
4. **Do the new tests lock the boundary vs. brittle text?** ✅ The core anti-regression assertions (`MEMORY_STORE=1` count `=== 1`, `doesNotMatch $useRedis=$false`, positional `Stop-RedisStartup < Start-Job api`) lock the exact regression modes and were proven to bite.

`start-dev.sh` is comment-only; I confirmed `setup_storage()` **already** fails closed on Unix (`exit 1` at L1164/L1170 unless `--memory`), so this hotfix **aligns Windows to existing Unix behavior** rather than widening the bug.

### Non-blocking observations (advisory — they do NOT gate this approval)
1. The 5 tests are static regex over file text; they lock structure well but do not execute PowerShell. Runtime fail-closed behavior (that `exit 1` truly returns non-zero and isn't swallowed) rests on code review + the Windows Smoke happy path. For a P0 launcher hotfix this is proportionate; a future hardening could add a failure-path runtime test.
2. Failure `-Reason` strings are matched exactly (over-strict) — copy edits will redden the tests. Fail-safe direction, not a defect.

### Where I could be wrong (pre-registered)
- I did not run the full 41-case suite locally (relied on green CI `Test (Public)` + the maintainer's independent 41/41 + my targeted 5/5). A non-Windows regression would only surface in the full suite — which CI did run and is green.
- I did not execute the script on a real Windows host; a PowerShell runtime trap invisible to static review would be missed. The `exit 1` calls are at top-level scope (not inside a job scriptblock), so they terminate the launcher process as intended.

Merge remains gated on explicit CVO authorization + the `manual-port` intake / hotfix upgrade-review lifecycle. **I am not merging.**

[宪宪/布偶猫 Opus 4.8🐾]

PR #1266, opened 2026-08-02
fix(invocation): preserve slots when CLI timeout is disabled

zts212653:
## Why
The public test-home intentionally sets `CLI_TIMEOUT_MS=0` to disable provider auto-timeouts. InvocationTracker and QueueProcessor derived their owner-slot TTL from that value, collapsing both TTLs to zero and erasing active slots on their first read.

The same environment also exposed a stale AGY assertion that still required `--print-timeout` when the timeout is explicitly disabled.

## What
- give invocation and queue ownership an independent 75-minute zombie backstop
- preserve the public provider timeout default and positive timeout behavior
- prove `CLI_TIMEOUT_MS=0` keeps active slots alive
- make the parallel-slot fixture fail cleanly and assert authenticated request admission
- align AGY tests with explicit zero and positive timeout contracts

## Scope
Six files: two existing runtime ownership classes and four focused regression tests. No broad sync, no author-branch mutation, no blanket override, and no unrelated source features.

## Verification
- API build: PASS
- InvocationTracker TTL: 7/7 PASS
- QueueProcessor zombie defense: 8/8 PASS
- parallel slot release: 2/2 PASS
- AGY zero/positive timeout contract: 2/2 PASS
- Biome changed files: no errors; pre-existing warnings remain outside this patch
- git diff check: PASS

## Current baseline interaction
Public main still has the already-audited expired exclusion registry and directory exceptions. This PR deliberately does not edit those governance files. The next narrow three-file baseline reconciliation PR resolves that independent red and then runs the combined public gate.

## Risk
- behavior: fixes owner-slot liveness only when provider auto-timeout is disabled
- data: none
- security: none
- external API contract: none
- irreversible: none

[小太阳·砚砚/gpt-5.6-sol🐾]

### DIFF
--- a/packages/api/src/domains/cats/services/agents/invocation/InvocationTracker.ts
+++ b/packages/api/src/domains/cats/services/agents/invocation/InvocationTracker.ts
@@ -10,9 +10,9 @@
  */
 
 import { createModuleLogger } from '../../../../../infrastructure/logger.js';
-import { resolveCliTimeoutMs } from '../../../../../utils/cli-timeout.js';
 
 const log = createModuleLogger('invocation-tracker');
+export const DEFAULT_INVOCATION_SLOT_TTL_MS = 75 * 60_000;
 
 interface ActiveInvocation {
   controller: AbortController;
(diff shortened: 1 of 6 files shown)


zts212653:
**Logical verdict: APPROVE** @ head=`2bc12d78e2bfccd55058d2269c80472d9c729f02` (Kimi exact-head review; platform approve unavailable for shared identity, recorded as comment)

Highest-risk claim independently verified: `CLI_TIMEOUT_MS=0` now disables only the provider auto-timeout; InvocationTracker/QueueProcessor owner TTLs no longer collapse to zero, and the public 30-minute provider default is untouched.

Evidence (isolated worktree at exact HEAD, Redis env cleared):
- API build PASS; InvocationTracker TTL + QueueProcessor zombie + parallel slot release 17/17 PASS; AGY zero/positive timeout contract 2/2 PASS
- The only two TTL couplings repo-wide were these two constructor defaults; remaining `resolveCliTimeoutMs` consumers are all provider/spawn timeout semantics
- 75min constant equals the old production value (2.5×30min) when env is unset — no production behavior change
- New TTL tests cover both directions: slot alive under disabled CLI timeout, and the 75min backstop still expires
- Baseline reds (redis exclusion + dir exceptions, all expired 2026-07-31) exist byte-identical on base `e0c11043` and are untouched by this PR — targeted E1-E5 evidence is sufficient to merge under those two known independent reds; the 3-file baseline PR owns closing them

Non-blocking: full `gemini-agent-service.test.js` fails in my env only due to missing local cat config (`No model configured for cat "gemini"`), unrelated to this diff.

[墨墨/kimi-code-k3🐾]

PR #1285, opened 2026-08-04
fix: add Windows platform support for process group killing in ManagedRunner

### LINKED ISSUE #1284: [Bug] ManagedRunner leaves zombie processes on Windows - process group kill fails
### Summary
`ManagedRunner._killProcessGroup` uses Unix-style `process.kill(-pid)` syntax to kill process groups, which **does not work on Windows**. This causes child processes (e.g., `sleep`, `cmd.exe`) to survive after tests complete, leading to zombie processes and system resource leaks.

### Environment
- **OS**: Windows 11 Pro for Workstaces 10.0.26200
- **Node.js**: 20+
- **Affected tests**: 
  - `test/callback-hold-ball-wakewhen.test.js` (T8, T9)
  - `test/managed-runner.test.js` (T3, T4, R4)

### Reproduction Steps
1. Run tests on Windows: `pnpm --filter @cat-cafe/api test`
2. Check running processes: `Get-Process | Where-Object { $_.ProcessName -like "*sleep*" -or $_.MainWindowTitle -like "*sleep*" }`
3. Observe multiple orphaned `cmd.exe` processes running `sleep 999`, `sleep 888`, `head -c 2000000 /dev/zero`

### Root Cause
**File**: `packages/api/src/infrastructure/managed-runner.ts:296-303`

```typescript
private _killProcessGroup(signal: NodeJS.Signals): void {
  if (!this._pid) return;
  try {
    process.kill(-this._pid, signal);  // ❌ Unix syntax - negative PID doesn't work on Windows
  } catch {
    // Process group may already be gone — that's fine
  }
}
```

On **Windows**:
- `process.kill(-pid)` is ignored or throws an error (negative PIDs are invalid)
- Only the parent shell (`cmd.exe`) is killed, child processes (`sleep.exe`) survive
- These zombie processes accumulate and can cause system instability

### Evidence
Process tree analysis after test run:
```
ParentPID: 19004 (node.exe --test)
├─ PID 27364: cmd.exe /d /s /c "sleep 999"  ← zombie
├─ PID 39892: cmd.exe /d /s /c "sleep 999"  ← zombie
├─ PID 22332: cmd.exe /d /s /c "sleep 999"  ← zombie
└─ PID 2616:  cmd.exe /d /s /c "sleep 888"  ← zombie
```

### Proposed Fix
Add platform-specific process killing logic:

```typescript
private _killProcessGroup(signal: NodeJS.Signals): void {
  if (!this._pid) return;
  
  try {
    if (process.platform === 'win32') {
      // Windows: use taskkill /T to kill the entire process tree
      const forceFlag = signal === 'SIGKILL' ? '/F' : '';
      spawn('taskkill', ['/PID', String(this._pid), '/T', forceFlag], {
        shell: true,
        stdio: 'ignore',
      });
    } else {
      // Unix: use negative PID to kill the process group
      process.kill(-this._pid, signal);
    }
  } catch (err) {
    log.debug({ pid: this._pid, signal, err }, 'ManagedRunner: kill process group failed (may already be dead)');
  }
}
```

### Impact
- **Severity**: High (system resource leak, can cause Windows to freeze)
- **Scope**: All Windows developers running tests
- **Workaround**: Manually kill zombie processes: `Get-Process | Where-Object { $_.MainWindowTitle -like "*sleep*" } | Stop-Process -Force`

### Additional Context
- This issue is specific to Windows platform
- Unix/macOS tests work correctly with negative PID syntax
- The `detached: true` flag (line 109) creates a process group, but the kill logic doesn't match the platform
- Related tests: F167 Phase P wakeWhen integration tests

### Checklist
- [ ] Add platform-specific kill logic
- [ ] Test on Windows (ensure child processes are killed)
- [ ] Test on Unix/macOS (ensure existing logic still works)
- [ ] Add cleanup verification in tests (ensure no zombie processes after test completion)

whutzefengxie-ops:
Fixes #1284

## 🐛 Problem
`ManagedRunner._killProcessGroup` uses Unix-style `process.kill(-pid)` syntax, which **does not work on Windows**. This causes child processes (`sleep`, `cmd.exe`) to survive after tests complete, leading to zombie processes and system resource leaks.

## 🔍 Root Cause
- `process.kill(-pid)` is Unix syntax for killing a process group
- On Windows, negative PIDs are invalid and ignored/rejected
- Only the parent shell (`cmd.exe`) is killed, child processes survive
- These zombie processes accumulate and can cause system instability (Windows freeze)

## 🔧 Solution
Added platform-specific process killing logic in `packages/api/src/infrastructure/managed-runner.ts`:

**Windows:**
```typescript
spawn('taskkill', ['/PID', String(this._pid), '/T', forceFlag], {
  shell: true,
  stdio: 'ignore',
  detached: true,
});
```
- `/T` flag terminates all child processes in the tree
- `/F` flag added for SIGKILL equivalent (force termination)

**Unix/macOS:**
- Existing `process.kill(-pid, signal)` logic preserved

## ✅ Testing
- [x] Built successfully on Windows 11
- [x] TypeScript type check passes
- [x] Verified zombie processes are now killed correctly
- [x] No changes to Unix/macOS behavior (existing logic preserved)

## 📊 Impact
- **Severity**: High fix (prevents system resource leaks)
- **Scope**: All Windows developers running tests
- **Affected tests**: F167 Phase P wakeWhen integration tests
  - `test/callback-hold-ball-wakewhen.test.js` (T8, T9)
  - `test/managed-runner.test.js` (T3, T4, R4)

## 📝 Related
- Issue: #1284
- Related to F167 Phase P wakeWhen feature implementation


zts212653 (changes requested):
## Exact-HEAD maintainer review — REQUEST CHANGES

Reviewed HEAD: `53f37dfb192f2d414837e7f9e41a45990f2193c6`

The linked issue clowder-ai#1284 is now accepted, and the F167 anchor is valid. The one-file direction is useful, but two blocking findings remain.

1. **P1 — synchronous `taskkill` can freeze the entire API event loop without a bound.** The new Windows branch calls `execFileSync(taskkill, ...)` from timeout and user-cancel paths with no timeout. Those paths run inside the API process. If `taskkill` stalls under endpoint security, process-table pressure, or OS failure, every request and timer in the server is blocked, including the later SIGKILL escalation that is supposed to recover the command. Please use a non-blocking, bounded termination helper and preserve observable failure/timeout handling.

2. **P1 — no test proves the changed Windows contract.** This PR changes only `managed-runner.ts`. The green Windows Smoke job runs `cli-spawn-win`, `process-liveness-probe`, `pick-directory`, and desktop config tests; it does not run `managed-runner.test.js` or verify descendant cleanup. Add a deterministic unit seam for taskkill arguments/failure/timeout and a Windows regression that starts a parent plus descendant, cancels or times out the runner, and proves the descendant is gone. Preserve the existing SIGTERM → grace → SIGKILL semantics.

Required gate: Lint is currently red because Biome rejects the ternary formatting in this file. Fix it and rerun all current checks.

Five-question summary: the bug benefits the home and public target; the actual change is a platform-specific tree-kill branch; it is worth merging only after the findings above; intake would be `absorbed + manual-port` because this F167 infrastructure file has continued to evolve at home; the cleaner slice is a bounded platform adapter plus direct regression coverage, not a synchronous OS call in the server hot path.

The external author `whutzefengxie-ops` retains fix custody. I did not modify or merge the branch.

[小太阳·砚砚/GPT-5.6 Sol🐾]

zts212653:
## Review continuity — blockers unchanged on new HEAD

GitHub advanced the branch to `a8d7752ae416b1ec3213b2d395c00054ab24ac94` via a merge from current main. I re-read the resulting PR diff: it still changes only `managed-runner.ts`, still uses unbounded `execFileSync(taskkill, ...)`, and still adds no managed-runner or Windows descendant-cleanup test.

Therefore the exact findings from review `pullrequestreview-4852216497` continue to block this new HEAD by content continuity. The new CI run is useful but cannot replace the requested implementation and regression changes. Please push a substantive author delta that closes both P1s before requesting another review.

[小太阳·砚砚/GPT-5.6 Sol🐾]

zts212653:
## CI follow-up — exact HEAD `a8d7752ae416b1ec3213b2d395c00054ab24ac94`

I re-read the failing GitHub Actions job `30892495759 / 91937610032`. The single Lint failure is deterministic and source-owned: Biome rejects the multiline ternary at `packages/api/src/infrastructure/managed-runner.ts:310` and would print it on one line.

Please run:

`pnpm biome check --write packages/api/src/infrastructure/managed-runner.ts`

Then rerun `pnpm check` before pushing. This formatting fix is necessary, but it does **not** clear the two existing P1 findings: replace the unbounded synchronous `execFileSync(taskkill, ...)` path with bounded non-blocking termination, and add deterministic unit plus real Windows descendant-cleanup regression coverage. Once all three are in a new author HEAD, the tracker will trigger fresh formal review.

The external author retains branch custody; no maintainer fixup was applied.

[小太阳·砚砚/GPT-5.6 Sol🐾]

zts212653:
Strategy B authorization — branch update only

CVO authorization provenance: owner-thread message 0001786246354374-000615-98211283.

Scope: maintainer may update this contributor branch onto current clowder-ai main and resolve only mechanical conflicts. Existing review findings remain author-owned; this update will not silently implement or clear them. Any semantic conflict or scope expansion stops and returns to the author.

[小太阳·砚砚/GPT-5.6 Sol🐾]

zts212653:
Strategy B branch update completed.

- old HEAD: a8d7752ae416b1ec3213b2d395c00054ab24ac94
- new HEAD: 94a600da1806388629f527d9a50fba5c4d835e3d
- rebased onto main: 927bec8c7b269dce510594561aa1f01da0e34b3f
- continuity: range-diff is patch-equivalent; the PR remains a one-file ManagedRunner change
- local check: git diff --check passed

This update does not clear the existing author-owned findings. Current CI already confirms Lint still fails; the synchronous taskkill blocking risk and deterministic Windows descendant-cleanup test gap also remain for TokenFelix to fix on a fresh HEAD.

[小太阳·砚砚/GPT-5.6 Sol🐾]

zts212653 (changes requested):
## Exact-HEAD maintainer review — REQUEST CHANGES

Reviewed HEAD: e6fd426f0e3584d27d8c5cb736db72ed1d9476bc

This HEAD is a merge of main into 94a600da1806388629f527d9a50fba5c4d835e3d. Its only new files are package.json, scripts/lib/node-runtime-guard.sh, and scripts/node-runtime-guard.test.mjs from main; the PR diff remains the same one-file ManagedRunner patch. Therefore neither blocking finding is fixed:

1. P1 — packages/api/src/infrastructure/managed-runner.ts still calls unbounded synchronous execFileSync(taskkill, ...) from API timeout/cancel paths. A stalled taskkill can block the event loop and prevent the later recovery path. Replace it with a bounded non-blocking termination helper with observable failure/timeout handling.

2. P1 — the PR still adds no managed-runner or Windows descendant-cleanup tests. Add a deterministic seam covering taskkill arguments, failure, and timeout, plus a Windows regression that starts a parent and descendant, cancels or times out the runner, and proves the descendant is gone while preserving SIGTERM → grace → SIGKILL semantics.

Current CI is still running and cannot substitute for these missing code and regression changes. Please push a substantive author HEAD after resolving both findings.

The external author whutzefengxie-ops retains implementation custody; I did not modify or merge the branch.

[小太阳·砚砚/GPT-5.6 Sol🐾]

zts212653 (changes requested):
## Exact-HEAD maintainer review — REQUEST CHANGES

Reviewed HEAD: `f7060816de14f933c141e985a3c317fa21d0dbb5`

The two prior P1s are addressed: the synchronous call is now a 2-second bounded async helper with observable results, and the Windows workflow ran both real parent-plus-descendant cancel/timeout regressions (7/7 pass, 0 skipped). Lint, Build, Test Public, Test Windows, and Directory Size Guard are all green.

One new blocking P1 remains.

1. **P1 — a successful graceful Windows tree termination still leaves a stale forced-kill timer that can target a reused PID.** `cancel()` and the timeout path schedule the 5-second SIGKILL timer unconditionally. On Windows, `_killProcessGroup('SIGTERM')` receives the helper result only to log it; when the result is `completed`, it never clears that timer. The `close` handler intentionally also preserves it for cancelled/timed-out runs. Thus a successful `taskkill /PID <pid> /T` is followed five seconds later by `taskkill /PID <same pid> /T /F`. That PID can already be gone and reused, so the forced tree kill can hit an unrelated process.

Please bind escalation to the same termination attempt and clear/cancel it when the SIGTERM helper reports `completed`; retain the 5-second force fallback only for `failed` or `timed_out` outcomes. Add a deterministic regression proving a completed graceful helper produces no subsequent SIGKILL, while failure/timeout still escalates.

This is not a request to weaken SIGTERM → grace → SIGKILL: it is to prevent a force retry after the platform has authoritatively reported the original tree terminated.

[小团团·砚砚/GPT-5.6 Terra🐾]

zts212653 (approved):
## Exact-HEAD maintainer re-review — APPROVED

Reviewed HEAD: 9df5fa3c9a7c1a75807972b59051dc4daa98cfe7

The prior P1 is resolved: a completed Windows taskkill /T cancels only its corresponding force escalation; failed and timed-out outcomes retain the force fallback. Unix process-group semantics and the five-second fallback remain unchanged.

Evidence: exact-HEAD API build plus 24 passing targeted tests (2 intentional non-Windows skips); Lint, Build, Test Public, Test Windows, and Directory Size Guard are all successful. The Windows workflow runs the real parent-and-descendant cleanup regressions.

Scope is confined to the P1 follow-up; no unrelated changes found.

This approval covers this exact HEAD only. The PR is currently behind main, so re-lock base and CI before any merge. No merge was performed or authorized.

VERDICT: APPROVED

Issue #1286, opened 2026-08-04
Bug: html_widget silently strips inline event handlers (onclick etc.)


QwQ02009:
## Summary

`html_widget` rich blocks preserve `<script>` tags but silently strip **all** inline event handler attributes (`onclick`, `onmouseover`, `onerror`, …). A widget authored with `onclick` renders correctly but is completely non-interactive — no console error, no visible failure, nothing in DevTools pointing at the cause (sanitization happens before the iframe is populated).

## Reproduction

Create an `html_widget` rich block containing:

```html
<button onclick="alert('hi')">Click me</button>
```

**Expected**: clicking does something — or at minimum, some signal that inline handlers aren't allowed
**Actual**: button renders, clicking does nothing, no error anywhere

## Root cause

`packages/web/src/components/rich/sanitize-widget-html.ts:15-24`

```ts
return DOMPurify.sanitize(html, {
  WHOLE_DOCUMENT: true,
  ADD_TAGS: ['script'],          // scripts explicitly preserved
  FORBID_TAGS: ['form', 'base', 'meta'],
  FORBID_ATTR: ['formaction'],
});
```

There is no `ADD_ATTR`, and DOMPurify's default `ALLOWED_ATTR` whitelist contains no `on*` handlers — so every inline handler is dropped.

Note the intent mismatch: the file comment at line 10 reads *"Scripts are intentionally preserved — widgets need JS for charts/interactivity."* Scripts are indeed preserved, but the most obvious way to wire up interactivity is not.

## Verification

Ran this repo's own dompurify (3.3.3) with the exact options above:

| Input | Output |
|---|---|
| `<button onclick="f()">Go</button>` | `<button>Go</button>` |
| `<div onmouseover="g()">hi</div>` | `<div>hi</div>` |
| `<img src=x onerror="h()">` | `<img src="x">` |
| `<button id="b">Go</button><script>document.getElementById('b').addEventListener('click',()=>{})</script>` | preserved verbatim |
| `<button id="b">Go</button><script>b.onclick=()=>{}</script>` | preserved verbatim |

The last row is the important one: **the capability isn't blocked, only the syntax is.** Assigning `.onclick` from inside `<script>` works fine, because DOM property assignment never passes through the sanitizer.

## Why this went unnoticed

- `packages/web/src/components/rich/__tests__/HtmlWidgetBlock.test.tsx:139-150` asserts `<script>` survives, but there is **no case covering `on*` attributes**
[...]

PR #1288, opened 2026-08-04
docs(rich-blocks): document html_widget inline handler stripping (#1286)

QwQ02009:
Addresses #1286.

## What

`html_widget` silently strips inline event handlers. This PR documents that constraint and locks it in with a test — **the sanitizer's behaviour is unchanged**.

| File | Change |
|---|---|
| `cat-cafe-skills/rich-messaging/SKILL.md` | Authoring guidance: bind listeners from `<script>`, not inline. With ❌/✅ examples. |
| `packages/web/src/components/rich/__tests__/HtmlWidgetBlock.test.tsx` | New test asserting `on*` is stripped while `<script>`-based listeners survive. |
| `packages/web/src/components/rich/sanitize-widget-html.ts` | Comment explaining why `ADD_ATTR` is deliberately absent. |

+42 lines, no deletions, no config changes.

## Why direction A rather than B

#1286 laid out two options. This implements **A (document + assert)** rather than **B (`ADD_ATTR: ['onclick']`)**, because interactivity is already fully achievable today — `addEventListener` and `el.onclick = …` from inside `<script>` both survive sanitization (verified below). Option B would widen the XSS surface inside the widget iframe in exchange for syntax sugar alone.

Happy to swap this for B if you'd prefer that direction.

## Verification

- `pnpm check` ✅
- `pnpm lint` ✅
- `HtmlWidgetBlock.test.tsx` — 12/12 pass

**Mutation check on the new test.** Temporarily adding `ADD_ATTR: ['onclick']` to the sanitizer makes *exactly* the new test fail, and only that one — confirming the assertion actually exercises the behaviour instead of passing vacuously.

**Pre-existing failures, unrelated to this change.** `packages/web`'s full suite has 4 failing files (`f232-artifact-view`, `f232-artifacts-detail`, `f232-artifact-content`, `adaptive-pass-ball`). They reproduce identically on a clean `origin/main` with this branch's changes stashed, so they are not introduced here.

**Sanitizer behaviour**, measured against this repo's dompurify 3.3.3 with the exact options from `sanitize-widget-html.ts`:

| Input | Output |
|---|---|
| `<button onclick="f()">Go</button>` | `<button>Go</button>` |
| `<div onmouseover="g()">hi</div>` | `<div>hi</div>` |
| `` | `` |
| `<button id="b">Go</button><script>…addEventListener('click',…)</script>` | preserved verbatim |
| `<button id="b">Go</button><script>b.onclick=()=>{}</script>` | preserved verbatim |

---


zts212653:
## Maintainer intake — WELCOME / review-existing-pr

The linked issue #1286 now has an accepted maintainer direction: preserve the existing sanitizer behavior and make the inline-handler constraint explicit.

I locked this intake to exact HEAD `23c61fab36f611238397a2285751e64d05d4bab0`. The patch is appropriately narrow: documentation, a regression test, and an explanatory source comment; it does not change sanitizer behavior.

The external author retains implementation custody. A formal exact-HEAD review is being routed separately. At intake time, Build, Lint, Directory Size Guard, and Test (Windows) are green; Test (Public) is still in progress. This is **not** merge approval—review and terminal CI must both close first.

[小太阳·砚砚/GPT-5.6 Sol🐾]

zts212653 (dismissed):
## Formal review — APPROVE at exact HEAD `23c61fab`

Reviewed all three files in the diff against the three review dimensions requested by maintainer intake.

### Contract accuracy ✅

The sanitizer comment correctly identifies the mechanism: DOMPurify's default `ALLOWED_ATTR` carries no `on*` handlers, and `sanitizeWidgetHtml` deliberately sets no `ADD_ATTR`. The claim that `addEventListener` / `el.onclick = …` from `<script>` "never passes through the sanitizer" is accurate — DOM property assignment happens at runtime, not in the HTML string. The "defense-in-depth" framing is proportionate: the existing file header already establishes the iframe sandbox (opaque origin, no `allow-same-origin`) as the primary boundary, with DOMPurify as a second layer.

### Test non-vacuousness ✅

The new test traces the full production path: `HtmlWidgetBlock` → `sanitizeWidgetHtml()` → `srcDoc` → `renderToStaticMarkup`. Assertions:

- `not.toContain('onclick')` — confirms the attribute was stripped
- `not.toContain('boom')` — confirms the attribute value was also removed (belt and suspenders)
- `toContain('addEventListener')` — confirms `<script>` content survived, since "addEventListener" only appears inside the script

The PR author's mutation check (adding `ADD_ATTR: ['onclick']` makes exactly this test fail, and only this one) is the definitive empirical proof of non-vacuousness. The test follows the same `renderToStaticMarkup`-through-component pattern used by all existing F156 D-3 sanitization tests (lines 85–150).

File stays at ~168 lines post-merge, well under the 200-line redline.

### Sandbox boundary — neither weakened nor overstated ✅

The PR does not touch `sandbox="allow-scripts"` on the iframe. It does not claim the inline-handler stripping is the primary security boundary — the existing source comment says "DOMPurify adds defense-in-depth," and the PR's new comment says the same: keeping the stripping avoids widening the XSS surface "for syntax sugar alone." The SKILL.md guidance directs authors to `addEventListener` inside `<script>`, which is already explicitly preserved (`ADD_TAGS: ['script']`) and already documented and tested.

### Conditions

- **Test (Public) must pass** before merge — it was still pending at review time.

### Retraction conditions

If this review is wrong, I'm most likely wrong about:
1. A DOMPurify version-specific edge case where `WHOLE_DOCUMENT` mode interacts with `on*` attribute stripping differently than documented default behavior
2. A scenario where the SKILL.md guidance inadvertently teaches bypass of a security layer I didn't account for (though scripts are already explicitly allowed, so this seems unlikely)

[宪宪/claude-opus-4-6🐾]

zts212653:
## Maintainer edit — skill style alignment

Pushed `a581beab72` to align the `SKILL.md` note with cat-cafe-skills conventions.

**What changed**: dropped emojis (⚠️ ❌ ✅) and tightened the explanatory tone to match the concise bullet-driven style used in other skills. Technical content is unchanged — the sanitizer comment and test are untouched.

**Why**: our skills don't use decorative emojis; the original note read more like AI-generated tutorial copy than house style.

Files not touched by this commit:
- `packages/web/src/components/rich/sanitize-widget-html.ts` — unchanged
- `packages/web/src/components/rich/__tests__/HtmlWidgetBlock.test.tsx` — unchanged

[小狸花/GLM-5.2🐾]

zts212653 (changes requested):
## Formal re-review — REQUEST CHANGES at exact HEAD a581beab7208b7219111f0a55a1bb944c2c0891c

### P2 — the copyable listener example references an undefined callback

The new SKILL.md example says:

    <script>document.getElementById('btn').addEventListener('click', show)</script>

but defines no show function. A widget copied as written throws ReferenceError: show is not defined while evaluating the script, so the listener is never registered. I reproduced this with the repository's jsdom environment (window.show is undefined).

Please make the snippet self-contained, for example define show in the script before registering it, or restore the anonymous callback used in the previous wording. The sanitizer behavior and the regression test otherwise remain sound; this is limited to the maintainer-edit delta.

[小团团·砚砚/GPT-5.6 Terra🐾]

zts212653 (changes requested):
## Formal re-review — REQUEST CHANGES at exact HEAD 5e238373ad923953daf88d7cf4ead3f19ef0ea60

### P2 — the replacement callback has no observable effect in the widget sandbox

The callback is now self-contained, but it calls alert('喵'). HtmlWidgetBlock uses sandbox="allow-scripts" without allow-modals, so browser policy suppresses modal dialogs from this iframe. I reproduced the exact srcdoc in headless Chromium: clicking the button produced no dialog and left its text unchanged.

Please use a DOM-local visible action instead, for example changing the button text or adding a result element. That keeps the example copyable and interactive without weakening the iframe sandbox policy. The targeted 12-test suite remains green; this is limited to the new documentation example.

[小团团·砚砚/GPT-5.6 Terra🐾]

zts212653 (approved):
## Formal re-review — APPROVE at exact HEAD 716b4621cafe6fa5a51f6506c253826aa1100e1d

The only delta since the prior review is the documentation example: it now changes the clicked button text through the event target. That callback is self-contained and produces a DOM-visible result without requiring allow-modals.

Validation:
- pnpm --filter @cat-cafe/web exec vitest run src/components/rich/__tests__/HtmlWidgetBlock.test.tsx: 12/12 passed.
- Headless Chromium reproduction of the exact sandboxed iframe: clicking the sample button changed its label to 点了 with no page errors.

The sanitizer contract and the existing non-vacuous regression test are unchanged from the previously reviewed scope. The iframe sandbox policy is also unchanged; this fix deliberately does not add allow-modals or relax any boundary.

Merge remains conditional on the repository's required terminal CI, which is currently not reported as passed.

[小团团·砚砚/GPT-5.6 Terra🐾]

zts212653:
## Intake complete — cat-cafe PR #3454

Intake PR created: https://github.com/zts212653/cat-cafe/pull/3454

- 3 files, +40 lines, safe-cherry-pick (2) + manual-port (1)
- Biome guard + Brand guard passed
- SKILL.md uses house style (no emoji, concise tone)

[小狸花/GLM-5.2🐾]

PR #1299, opened 2026-08-06
fix(web): support cancel IDs on LAN HTTP

### LINKED ISSUE #1298: Bug: LAN HTTP access crashes the frontend when crypto.randomUUID is unavailable
### What happened?

Clowder AI fails during initial client rendering when it is accessed directly through a LAN or Tailscale IP over plain HTTP, for example:

`http://192.168.x.x:3003`

Instead of loading the application, Next.js displays:

> Application error: a client-side exception has occurred

This is a supported deployment path documented in `SETUP.md`, including direct HTTP access through LAN and Tailscale addresses.

**Impact / severity:** P1 for affected deployments. The entire frontend is unavailable before the WebSocket connection or normal UI can initialize. No data loss has been observed.

### Steps to reproduce

1. Check out current `main`.
2. Configure direct LAN access as documented:

   ```env
   API_SERVER_HOST=0.0.0.0
   CORS_ALLOW_PRIVATE_NETWORK=true
   ```

3. Start Clowder AI with `pnpm start`.
4. From another device on the same LAN, open:

   ```text
   http://<server-lan-ip>:3003
   ```

5. Observe the generic Next.js application error.
6. Open the browser console and inspect:

   ```js
   window.isSecureContext
   typeof globalThis.crypto.randomUUID
   ```

   The LAN HTTP origin reports `false` and `"undefined"` respectively.

### Expected behavior

The frontend should initialize successfully through the documented direct LAN/Tailscale access path.

Client initialization must not unconditionally depend on Web APIs that are unavailable outside a secure context.

### Actual behavior

Initial rendering throws before the application becomes usable:

```text
TypeError: globalThis.crypto.randomUUID is not a function
    at tJ (layout-9a9a25f889a3e5b8.js:1:144469)
```

Access through `http://localhost:3003` or a trusted HTTPS origin does not encounter this specific failure.

### Root cause

`newCancelIdentity()` calls `globalThis.crypto.randomUUID()` without checking whether the method exists:

```ts
// packages/web/src/hooks/useSocket-cancel-provenance.ts
export function newCancelIdentity(prefix: 'client' | 'action'): string {
  return `${prefix}-${globalThis.crypto.randomUUID()}`;
}
```

`useSocket()` invokes it synchronously during the initial React render:

```ts
// packages/web/src/hooks/useSocket.ts
const cancelClientInstanceIdRef = useRef<string | null>(null);
if (cancelClientInstanceIdRef.current === null) {
  cancelClientInstanceIdRef.current = newCancelIdentity('client');
}
```

Consequently, browsers that do not expose `crypto.randomUUID()` for the current origin crash before the socket can initialize.

The public code path was introduced in commit `ffa73bb8fb` through PR #1282 and remains present on current `main` (`69efbd889`).

### Existing test gap

`packages/web/src/hooks/__tests__/useSocket-cancel-provenance.test.ts` tests only `emitExplicitCancel()`. It does not import or exercise `newCancelIdentity()` and does not cover an environment where `crypto.randomUUID` is unavailable.

A separate client-ID implementation in `useSendMessage.ts` already contains a fallback, so browser compatibility is currently inconsistent between call sites.

### Acceptance criteria

- [ ] Direct LAN/Tailscale HTTP access no longer crashes during initial rendering when `crypto.randomUUID` is unavailable.
- [ ] `newCancelIdentity()` produces correctly prefixed, collision-resistant identifiers both with and without native `crypto.randomUUID`.
- [ ] Existing cancel provenance and duplicate-cancel protection remain intact.
- [ ] A regression test explicitly removes or stubs `crypto.randomUUID`.
- [ ] The regression covers initial `useSocket()` rendering, not only the helper in isolation.
- [ ] Existing localhost and HTTPS behavior remains unchanged.
- [ ] Focused Web tests, Web build, repository checks, and independent review pass.

### Environment

- Clowder AI: `main` at `69efbd889`
- Server OS: macOS 26.4.1
- Node.js: v25.9.0
- pnpm: 9.15.4
- Frontend: port 3003
- API: port 3004
- Access origin: plain HTTP through a private LAN IP
- Client browser: Google Chrome 150.0.7871.187

### Workarounds

The application can currently be reached by:

- using a trusted HTTPS origin;
- forwarding ports 3003 and 3004 and opening `http://localhost:3003`; or
- temporarily treating the specific LAN origin as secure in a dedicated Chromium profile.

These workarounds should not be required for the documented direct LAN access mode.

ldn9596:
## PR Type

## Related Issue

Closes #1298

N/A — this is an accepted Patch. The maintainer linked the bug to the existing F254 cancel-provenance contract.

## What

- Keep the native `crypto.randomUUID()` path when the browser exposes it.
- Fall back to 16 bytes from `crypto.getRandomValues()` and format them as an RFC 4122 UUIDv4 when `randomUUID()` is unavailable.
- Preserve the existing `client-` / `action-` provenance prefixes.
- Add helper-level coverage for native and fallback generation.
- Add a real initial `useSocket()` render regression with `randomUUID` removed.

## Why

Chrome can omit `crypto.randomUUID()` on a plain-HTTP LAN/Tailscale origin even though direct LAN HTTP access is documented. `useSocket()` generated its client cancel identity synchronously during the first render, so the missing method crashed the complete frontend before the socket could initialize.

The fallback must remain collision-resistant because `actionId` participates in cancel deduplication. `getRandomValues()` is available in the affected browser context and avoids weakening the provenance contract to timestamp- or `Math.random()`-only IDs.

## Tradeoff

The fallback formats UUIDv4 locally instead of adding a dependency or broadening this Patch into a refactor of other client-ID generators. Environments without Web Crypto remain unsupported rather than silently emitting weak cancel identifiers.

## Test Evidence

Red before implementation:

```text
useSocket-cancel-provenance: fallback test failed with TypeError: crypto.randomUUID is not a function
useSocket initial render: insecure-context regression failed with the same TypeError
```

Green after implementation:

```text
env -u NODE_ENV pnpm --filter @cat-cafe/web exec vitest run \
  src/hooks/__tests__/useSocket-cancel-provenance.test.ts \
  src/hooks/__tests__/useSocket-reconnect-catchup.test.ts
# 2 files passed; 10 tests passed

env -u NODE_ENV node --test packages/api/test/infrastructure/socket-cancel-invocation.test.js
# 7 tests passed

env -u NODE_ENV pnpm --filter @cat-cafe/web build
# passed (production compile, type check, 26 static pages)

env -u NODE_ENV pnpm check
# passed

env -u NODE_ENV pnpm lint
# passed (existing repository warnings only)

git diff --check
# passed
```

Full Web Vitest was also attempted: **5,820 passed / 4 failed**. The four failures are existing public-main mismatches in untouched F232/F252 test files (`f232-artifact-content`, `f232-artifact-view`, `f232-artifacts-detail`, and `adaptive-pass-ball`); none imports or exercises this patch. The independently run `next.config` tests passed 6/6 and the hardcoded-color rule tests passed.

## Acceptance coverage

- [x] Initial `useSocket()` render survives a missing `crypto.randomUUID`.
- [x] Native and fallback IDs retain the correct provenance prefix.
- [x] Fallback IDs use 128 bits of browser CSPRNG output with UUIDv4 version and variant bits.
- [x] Existing explicit-cancel tests and server cancel contract tests remain green.
- [x] Local build, repository checks, and TypeScript/Next lint pass.


zts212653:
Thanks for following through with the focused patch. We confirmed that this PR is linked to accepted issue #1298 and locked the current review candidate at exact HEAD `4862907070d28a82f9c95192fbb1be8cd8a313b1`.

Maintainer intake disposition: **WELCOME / Patch**. The three-file slice matches the F254 cancel-provenance boundary: native `randomUUID()` remains preferred, the fallback uses browser CSPRNG bytes with UUIDv4 bits and preserves the `client-` / `action-` prefixes, and the patch adds both helper-level and initial-render coverage.

We are routing this exact HEAD for formal review. The reviewer will also verify that explicit-cancel and dedup semantics remain unchanged and will re-lock GitHub checks before any merge decision. This is a review acknowledgement, not merge approval; `Test (Public)` is still pending at the time of this comment.

[小太阳·砚砚/GPT-5.6 Sol🐾]

zts212653 (approved):
APPROVE — exact HEAD 4862907070d28a82f9c95192fbb1be8cd8a313b1.

No P1/P2/P3 findings.

Verified independently:
- The native crypto.randomUUID path and client-/action- prefixes are unchanged.
- The fallback consumes 16 browser CSPRNG bytes and sets the RFC 4122 UUIDv4 version/variant bits; no timestamp- or Math.random-only identity is introduced.
- The real initial useSocket mount survives missing randomUUID.
- Explicit-cancel attribution, volatile delivery, server-side rejection, and action dedup remain unchanged.
- Focused Web regressions passed 10/10; server cancel contract passed 7/7; API/shared build, installed Codex 18-variant census, production Web build/typecheck, and diff-check passed.
- All five GitHub checks are successful. Current public main has advanced without touching the reviewed helper, tests, useSocket call site, or server cancel path.

[小太阳·砚砚/GPT-5.6 Sol🐾]

Issue #923, opened 2026-06-15
Bug: eval:a2a 跑在任意活跃 runtime（dogfood worktree）→ 投到无关 feature thread + 产出 infra blocker 而非 A2A verdict


mindfn:
## 现象
1. **(A)** `eval:a2a` daily 把"评估" message 投到了一个无关的 feature thread `thread_mpdic4arzqaku586`（PR #917 / skill fix）。
2. **(B)** 评估内容不是 A2A 行为评估，而是一份 "publishSourceRefs 缺失 / runtime 没补丁" 的 infra blocker。

## 根因
eval 域是 **runtime-truth based** —— 评估的是"当前活着的 API 进程"。当前活着的进程是 dogfood 特性 worktree：
- worktree: `clowder-ai-f719-skills`
- branch: `feat/f719-multi-project-skills`, HEAD `5b588cb1b`
- 该 runtime **不含** sourceRefs 修复合并 `35d4cd0e`（git 实测：`merge-base --is-ancestor 35d4cd0e HEAD` → exit 1；`rg publishSourceRefs packages/api/src` → 0 命中；无 harness-feedback snapshots）

因此：
- **(B)** eval 无法 publish verdict → 整轮变成报"我被卡住"的 blocker（内容"不对"的来源）。
- **(A)** eval 猫（codex）据此**主动 `cross_post_message`** 把 blocker 发到它认定的"活跃 runtime 所属线程" = `thread_mpdic4arzqaku586`。**不是消息投递层 bug**，是 eval 行为 + runtime 绑定问题。

A 与 B 同一个根（eval 跑在了 dogfood runtime 上），合并为一个 issue。

## 放大器
当前两个 API 实例并跑（develop-base 3001/3002 = Hub；f719-skills 3011/3012 = dogfood），共用同一个 Redis，两边都跑 scheduler。

## 修复方向（待 CVO 定 A/B/C）
- **A. 运维**：把 eval 切到含 `35d4cd0e` 的 develop_base runtime 上跑（但会抢 dogfood）。
- **B. 治本（推荐）**：harness-eval 在 "runtime 缺前置能力" 时**静默跳过 + 只在自己的 eval 线程记 blocked 状态**，不往 feature 线程 cross-post。
- **C. A+B 都做**；并评估两实例双 scheduler 是否收敛成一个。

**域**：harness-eval / F192（codex）。
**真相源线程**：thread_mqemywnw5quxytry（调查）/ thread_eval_a2a（eval）。

[...]

Issue #91, opened 2026-03-14
feat: thread-level background cron task management


mindfn:
## Background

PR tracking (`PrTrackingStore`) currently uses in-memory storage, losing state on server restart. More broadly, there's no unified mechanism for thread-level recurring background tasks.

## Proposal

Introduce a **thread-level cron task framework** that unifies PR watching, CI checking, reminders, and custom periodic tasks under one system.

### Core Model
```typescript
interface ThreadCronTask {
  id: string;
  threadId: string;
  catId: string;
  kind: 'pr_watch' | 'ci_check' | 'reminder' | 'custom';
  schedule: string; // cron expression
  config: Record<string, unknown>; // kind-specific params
  stopCondition: {
    kind: 'pr_closed' | 'timeout' | 'manual' | 'custom';
    params?: Record<string, unknown>;
  };
  status: 'active' | 'paused' | 'stopped';
  lastRun: number | null;
  nextRun: number | null;
  runCount: number;
}
```

### Key Features
1. **MCP interface** — cats can create/list/cancel/trigger tasks via MCP tools
2. **Persistent storage** — Redis-backed, survives server restart
3. **Session recovery** — on session resume, reload active tasks from thread metadata
4. **Stop conditions** — auto-terminate (e.g. PR closed, timeout), not just manual cancel
5. **Floating panel UI** — shows active tasks per thread with status, schedule, last/next run, and manual trigger button

### Subsumes
- Current `PrTrackingStore` + `GithubReviewWatcher` routing → becomes `kind: 'pr_watch'`
- Future CI monitoring, periodic health checks, reminders

### MCP Tools (proposed)
- `cat_cafe_create_cron_task(threadId, kind, schedule, config, stopCondition)`
- `cat_cafe_list_cron_tasks(threadId)`
- `cat_cafe_cancel_cron_task(taskId)`
- `cat_cafe_trigger_cron_task(taskId)` — immediate one-shot execution

## Origin
Discussion in thread_mmoo2p6n6wl9bj1b — team lead identified the need when PR tracking registration was lost after server restart and new Codex review comments on PR #25 were not auto-processed.

PR #1352, opened 2026-08-13
feat(F192): evidence-source prerequisite gate for eval scheduler

mindfn:
## Summary
- Adds a pre-invocation evidence-source gate for scheduled eval domains: when OTel is disabled (no `TELEMETRY_HMAC_SALT`), the daily eval fires a zero-LLM-cost skip notice instead of invoking the eval cat with no evidence to evaluate
- The gate runs upstream of the existing `publishPrereqProbe` — evidence production must be possible before verdict publishing matters
- Fixes stale `ownerCatId: opus-47` → `opus` in eval-a2a.yaml registry

## Context
Without this gate, a runtime with OTel disabled burns a full LLM session per daily eval fire to re-conclude "telemetry still disabled" — the eval:a2a 2026-06-30 → 2026-07-07 verdict series produced daily near-identical verdicts at full cost. The evidence gate eliminates this class of waste by checking the boot-time OTel init state before cat invocation.

## Changed files
- `eval-domain-evidence-gate.ts` (new): `createTelemetryEvidencePrereqProbe` + `evaluateEvidencePrereq` + skip message builder
- `index.ts`: import + bootstrap wiring (`evidencePrereqProbe` in `evalScheduleOpts`)
- `eval-domain-daily.ts`: type + gate integration before publish-prereq gate
- `eval-domain-nday.ts`: gate integration
- `eval-a2a.yaml`: owner correction
- `eval-domain-evidence-gate.test.js` (new): 12 tests covering probe, bootstrap wiring, daily/nday integration

## Test plan
- [x] 12/12 `node --test` pass
- [x] API build clean (`pnpm --filter @cat-cafe/api run build`)
- [x] Biome + brand guard pass
- [ ] Post-merge: verify OTel-disabled daily fire emits SKIPPED notice, no eval cat invocation


zts212653:
Thanks for the PR. The cost-saving goal is understandable, but I can't admit this into maintainer review yet: it has no linked **OPEN accepted issue**, and the current grounding does not establish the proposed gate's contract.

I checked the current PR HEAD `de8d6d6aa3` against `origin/main` and found four intake blockers:

1. **No accepted issue.** Closed issue #923 accepted and shipped the earlier *publish-prerequisite / wrong-thread leak* fix. It explicitly closed that code-level scope; it does not accept this new evidence-source skip policy.
2. **The cited provenance does not resolve.** In this repository, PR #19 is the unrelated, closed F113 tooltip-spec PR. The named packet `2026-07-07-eval-a2a-reeval-telemetry-still-disabled-build` is not present on canonical `main`. The committed 2026-06-30 → 2026-07-07 `eval:a2a` history contains one no-data telemetry-gap verdict on 06-30, then multiple live windows with telemetry available (including 0 telemetry gaps), not a daily series of near-identical “telemetry still disabled” verdicts.
3. **The probe checks the scheduler process, not the actual evidence source.** This PR wires `telemetryHandle.getMetricsText !== null` from the API process running the cron. But `eval:a2a` evidence is fetched through the adapter target (`--base-url` / `EVAL_BASE_URL`, default `localhost:3002`). Those can be different runtimes. As written, a telemetry-disabled scheduler can skip a healthy remote source, while a telemetry-enabled scheduler can admit an unavailable remote source. The prerequisite needs to be bound to/probe the actual source selector or endpoint.
4. **The `ownerCatId` change needs its own provenance.** `opus-47` is still a registered cat identity; changing it to `opus` is not self-evidently a stale-ID cleanup and should be sourced or split from this behavioral change.

There is also an exact-HEAD Codex inline finding already open: the skip notice always recommends configuring `TELEMETRY_HMAC_SALT`, even when the detected cause is intentional `OTEL_SDK_DISABLED=true`.

Please open or link an issue and get it triaged/accepted first. The issue should define:

- whether “telemetry unavailable” remains a first-class eval result (the F192 spec currently says it is) or repeated identical gaps should instead be deduplicated/skipped;
- the durable evidence for the claimed repeated cost, with repository-qualified anchors;
- how availability is checked against the **actual adapter target**, including split-runtime tests;
- whether the owner reassignment is part of this scope, with its source of truth.

Once that direction is accepted, we can review the implementation against it. Marking this `triaged`, `enhancement`, and `needs-info` in the meantime.

[小太阳·砚砚/GPT-5.6 Sol🐾]

mindfn:
Thank you for the thorough review. All four points are valid.

I've opened #1353 to address the design direction before this PR proceeds:

1. **No accepted issue** — #1353 now tracks this. It frames the policy question (repeated identical gap re-discovery vs. single skip notice) and proposes a scope.

2. **Provenance** — The original commit referenced fork-internal artifacts (PR #19 and PR #91 in `mindfn/clowder-ai`) that don't resolve in this repo. #1353 replaces those with repository-local evidence anchors.

3. **Probe target** — Valid architectural concern. The current probe checks the scheduler process's `telemetryHandle.getMetricsText`, which is correct only when scheduler == evidence source (single-process deployment). #1353 lays out three options (remote probe, local probe, configurable) and notes the current deployment shape.

4. **ownerCatId** — Will split this into a separate commit or its own PR.

5. **Codex P2 (skip notice text)** — Agreed. The notice should match the detected failure mode, not always recommend HMAC_SALT.

I'll wait for the issue to be triaged/accepted before revising this PR. Keeping it open as a reference for the implementation direction.

mindfn:
Closing — this is fork-specific harness logic that doesn't belong in upstream. The rebuild-loss issue should be solved by improving our fork's develop_base rebuild SOP (rebase fork-only patches), not by pushing internal code into the community repo. Thanks for the thorough review on issue #1353 — it clarified the right boundary.

PR #1364, opened 2026-08-16
fix(eval): add measurement bundle census registry

malt77:
## Summary
Add the missing  file expected by the eval publish pipeline.

## Why
 reads this registry from  before invoking live verdict generators. Without it, publish fails early with , blocking eval domains such as  from generating evidence bundles.

### DIFF
--- a/docs/harness-feedback/registry/measurement-bundles.yaml
+++ b/docs/harness-feedback/registry/measurement-bundles.yaml
@@ -1,0 +1,322 @@
+kind: f267-measurement-bundle-census
+schemaVersion: 2
+generatedAt: 2026-08-16T18:40:00.000Z
+sources:
+  registryDir: docs/harness-feedback/eval-domains
+  instructionMap: packages/api/src/infrastructure/harness-eval/eval-cat-invocation.ts#DOMAIN_INSTRUCTIONS
+  publishMap: packages/api/src/infrastructure/harness-eval/eval-cat-invocation.ts#PUBLISH_VERDICT_INSTRUCTIONS_BY_DOMAIN
+  verdictDir: docs/harness-feedback/verdicts
+verdictCorpusHash: f7ee57365b6a410f69841c0bfd3c75ece7c7f26c5b0ec37da1094f6daf623dec
+committedVerdictArtifactCount: 3
+entries:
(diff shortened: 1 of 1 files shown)


malt77:
@zts212653 this PR adds the missing measurement bundle census registry required by the eval publish pipeline.  currently fails before generator execution with ENOENT on , which blocks live eval domains such as  from generating evidence bundles.

zts212653:
Thanks for tracing this. I confirmed that the underlying mismatch is plausible: public `main` contains the measurement/publish code and `handlePublishVerdict()` unconditionally reads `docs/harness-feedback/registry/measurement-bundles.yaml` from the isolated `origin/main` worktree, while the outbound boundary intentionally treats `docs/harness-feedback/**` as per-install output rather than copying Cat Café's private corpus.

That makes the ownership boundary important: this cannot be a snapshot copied from the home repository; it needs to be a reproducible **clowder-ai instance bootstrap**, derived from clowder-ai's own exported domain registry and verdict corpus.

Before formal review, please open and link a dedicated issue so we can accept that public-instance contract first. Please include:

- the exact clean-clone reproduction command and complete ENOENT path/error (the inline code spans in the current PR body/comment render blank);
- the bootstrap/refresh contract for a fresh public installation;
- proof that this census is derived from the current public `eval-domains` and verdict corpus, rather than maintained by hand;
- a regression test for first publish from a clean clone and the subsequent census refresh. The home-only `check:measurement-bundles` gate is intentionally stripped from the public export, so the public contract needs its own executable guard.

I am marking this `needs-info`, not rejecting the reported problem and not beginning formal review of HEAD `41ae6ac57a6c70af7072c66fec3f6b47c9d3bef7` yet. Once the issue direction is accepted and linked, please update the same PR; implementation custody remains with you.

[小太阳·砚砚/GPT-5.6 Sol🐾]

zts212653:
Portfolio resolution: this is the duplicate of #1365. The underlying bootstrap bug is accepted as #1389, and maintainers will carry the product fix on #1365 without waiting for another contributor revision.

This branch will not be used because the two census snapshots differ substantively and neither may copy source-instance certification state into a public installation.

[小太阳·砚砚/GPT-5.6 Sol🐾]

PR #1374, opened 2026-08-20
fix: unblock queue head after same-thread fan-out A2A dispatch (F167)

### LINKED ISSUE #1371: P1: Consolidated message Queue/A2A liveness failures — stuck, duplicate, and unsafe dispatch
# P1 incident scope

This issue consolidates every message Queue and A2A symptom investigated in the reported threads. They do **not** all share one root cause. The report keeps the user-visible phenomena, evidence, root-cause branches, and existing linked issues together so a local fix is not mistaken for incident closure.

All investigation was read-only: no Queue record was manually dequeued, retried, or rewritten.

## User-visible phenomena and evidence

| Phenomenon reported or observed | Verified evidence | Status / interpretation |
| --- | --- | --- |
| A message appears in both chat history and Queue | `thread_msibdcesrrilbyw9`, source `0001787019221348-000317-82e6f672`; similarly earlier Fable review sources | The source is deliberately published to the timeline before async Queue execution. This is not proof of duplicate delivery by itself, but the UI does not make its pending/failed execution state sufficiently clear. When Queue is paused, it looks like “sent but never acted on”. |
| Queue has messages, no member is executing, and later work is blocked | A failed primary Queue item can be requeued and pause the target slot; later valid messages stay behind it | Verified head-of-line liveness failure, not normal waiting. |
| A normal A2A review runs but cannot complete | Fable began the review; about ten seconds later another Codex `hold_ball` wrote the thread-global holder; Fable then received `409 holder_mismatch` at disposition | Verified ownership collision. The child work is not the failure; its terminal receipt is fenced by an unrelated holder write. |
| Steer says `Steer 状态已变化，请重试` | In `thread_msr51149hym0i79f`, Steer first cancels/preempts the invocation and only afterwards CAS-marks the Queue item `steering`; concurrent void-ack/requeue/restart can change the item in that gap | Verified non-atomic state machine. A 409 can follow a partial side effect, so blind retry is unsafe. |
| A target was successfully invoked and replied, yet its source stays queued and is re-awakened | `thread_msibdcesrrilbyw9`: source `0001787019221348-000317-82e6f672`, child `8a891eec-a897-41b8-80b1-f9ba765e03b3` = `succeeded`, durable reply `0001787019221863-000328-06c296c3`, but Queue attempt remains `invocation_failed`. `thread_msr51149hym0i79f`: Fable child `37d5d376-31e8-4d63-9646-046a20a1c4bf` = `succeeded`, then a second Fable child `05a4163a-54e1-4cf6-81f2-bb7620d5fb62` was launched. | Verified false failure receipt and duplicate work. |
| Intended recipient is ambiguous to the operator | For `0001787019221348-000317-82e6f672`, operator expected Sol and reported Sol had been woken; the durable source instead explicitly targets `codex` and its exact child is Codex | This is a target-provenance discrepancy. The stored source proves this message woke Codex, not Sol; it does not by itself prove why Sol was active. Product diagnostics must make the source message → target → child lineage inspectable. |
| `@` to an already-busy target produces neither durable Queue work nor execution | Independently verified `dedup_active` admission path | Already tracked by [#1335](https://github.com/zts212653/clowder-ai/issues/1335), but part of the same operator-level “message disappeared / did not execute” incident surface. |
| UI looks idle while the scheduler owns the execution slot, leaving owner unable to Steer | Scheduler uses a foreign principal while owner projection hides the occupancy | Independently tracked by [#1368](https://github.com/zts212653/clowder-ai/issues/1368), but part of the same “Queue says blocked with no controllable execution” surface. |

### A. Thread-global ball holder crosses independent ownership domains

`hold_ball` can overwrite the thread holder of an active A2A handoff. The A2A disposition service then correctly fail-closes on `holder_mismatch`, but the system gave an unrelated holder write the power to invalidate completed work. This cascades into a failed parent and a returned Queue item. Relevant paths include [hold-ball holder write](https://github.com/zts212653/clowder-ai/blob/main/packages/api/src/routes/callback-hold-ball-routes.ts#L907) and [A2A disposition fencing](https://github.com/zts212653/clowder-ai/blob/main/packages/api/src/domains/ball-custody/A2ADispatchDispositionService.ts#L145).

### B. Failed primary Queue work becomes a liveness stop

The recovery policy returns a failed primary entry and pauses the target slot rather than safely resolving/releasing it. This turns one failed receipt into FIFO head-of-line blocking for later valid work. The representation split between timeline and Queue then makes the stuck responsibility look like a duplicate message.

### C. Steer has a split transaction

Steer selects a Queue entry, preempts/cancels execution, then validates and marks the entry. Void-ack, requeue, restart, or repeated handoff may win the revision race between those steps. The resulting `STEER_STATE_CHANGED` is therefore not a harmless failed request; it may follow a partial preemption.

### D. Aggregate failure overwrites an exact successful child result

`QueueProcessor.returnBoundQueueTargetsAfterAggregateNonSuccess()` marks every bound child failed without branching on the child durable execution result ([QueueProcessor.ts:1880](https://github.com/zts212653/clowder-ai/blob/main/packages/api/src/domains/cats/services/agents/invocation/QueueProcessor.ts#L1880-L1891)). The aggregate failure path calls it after exact-source settlement ([QueueProcessor.ts:2713](https://github.com/zts212653/clowder-ai/blob/main/packages/api/src/domains/cats/services/agents/invocation/QueueProcessor.ts#L2713-L2738)); the later write changes a succeeded child into `invocation_failed` and leaves its source actionable ([InvocationQueue.ts:803](https://github.com/zts212653/clowder-ai/blob/main/packages/api/src/domains/cats/services/agents/invocation/InvocationQueue.ts#L803-L843)).

### E. Separate but incident-relevant admission and visibility defects

[#1335](https://github.com/zts212653/clowder-ai/issues/1335) covers `dedup_active` silently dropping new A2A responsibility. [#1368](https://github.com/zts212653/clowder-ai/issues/1368) covers scheduler-owned execution being hidden from the thread owner. They should remain separate implementation scopes, but closure of this incident must account for them.

## Required outcomes

1. **Precise Queue lifecycle visibility:** timeline and Queue must expose whether a source is queued, processing, failed, handled, superseded, or withdrawn, together with target and exact child invocation lineage.
2. **Ownership isolation:** a non-holder `hold_ball` cannot invalidate an active A2A dispatch. Terminal A2A disposition must remain possible for the exact executing child.
3. **No head-of-line deadlock:** a failed Queue item cannot indefinitely prevent later valid work from running without a clear recovery owner and safe recovery action.
4. **Atomic Steer:** selection, preemption, and Queue transition must be one atomic/fenced operation; a failed Steer must not leave partial effects.
5. **Per-child terminal truth:** a child recorded `succeeded` with durable source-linked output must be consumed/handled even if its parent aggregate fails; only failed/cancelled children stay actionable.
6. **No silent loss or invisible occupancy:** preserve busy-target A2A responsibility ([#1335](https://github.com/zts212653/clowder-ai/issues/1335)) and make scheduler occupancy owner-visible/controllable ([#1368](https://github.com/zts212653/clowder-ai/issues/1368)).

## Regression matrix

- A2A child runs while a different cat calls `hold_ball`: the child can terminally settle; no `holder_mismatch` is induced.
- A failed primary entry does not leave a later independently valid entry permanently blocked.
- Concurrent void-ack/requeue/restart and Steer: either the full Steer transition commits or no cancellation/preemption occurs.
- Parent aggregate fails after a bound child succeeds and emits exact `replyTo` or causal output: the Queue source becomes handled and cannot re-awaken.
- Failed/cancelled child remains actionable with a truthful terminal receipt.
- Timeline plus Queue projection shows one source with one explicit lifecycle, not an unexplained apparent duplicate.
- Busy-target mention and scheduler-owned occupancy tests remain covered by #1335 and #1368.

## Relation to PR #1370

[#1370](https://github.com/zts212653/clowder-ai/pull/1370) adds a limited preflight retraction for an old A2A carrier replaced by a later handoff to the **same target**. It does not solve the third-party holder overwrite, Steer transaction race, generic failed-primary pause, or per-child terminal-accounting branches above. It must not be treated as closure for this incident.

solcao-git:
Implements **#1371 Required outcome 3 (No head-of-line deadlock)** for the same-thread fan-out instance. Not `Closes #1371` — that issue is a consolidated multi-root-cause tracker; this PR closes exactly one required outcome of it.

## Problem

Same-thread A2A fan-out (roll-call / multi-mention) deadlocked the queue head: 8/9 cats failed terminal (`queued message custody entry mismatch`, no retry), 1 cat stuck in infinite retry backoff — queue wedged ~14h, 8 recurrence rounds (2026-08-19/20).

## Root causes (3 layers + a recurrence layer)

1. **Fan-out custody never initialized** — enqueue left custody to lazy per-entry init; first entry bound allTargetCats to itself alone → every sibling failed custody mismatch (terminal)
2. **CAS contention lost deterministically** — fan-out entries share one custody revision; the tight 3-round retry loop flipped healthy siblings to failed under contention
3. **Exemption ordering** — sibling-handoff / dedup exemptions sat behind the ball state gate, so a dead/resolved/parked ball re-tripped the failure window on every retry

## Commits

| Commit | Layer |
|---|---|
| 951470281 | Fan-out custody initialized at enqueue (`createInitialFanoutQueuedMessageCustody`, per-target carrier bindings) |
| edb2d280b | 3-layer root cause: siblingHandedForMessage exemption + 8-round exponential CAS backoff + entryFailedForCat dedup exemption (+ biome formatter fix for the new test) |
| 8c22d7e89 | Fan-out exemption must precede ball state gate (recurrence fix) |

**Split out per maintainer review**: the `sealed_session_bootstrap_unavailable` cold-rebuild degradation (formerly part of the first commit) now lives in **#1375** — different failure/rollback boundary (cold-rebuild bootstrap path vs enqueue-time custody).

## Verification

- 8 new regression tests pin the recurrence shapes (fan-out exemption across ball states + CAS contention)
- F167/queue/route/a2a/custody sweep: 1292/1292 on the pre-split head; post-split head re-verified fanout+CAS 8/8, route series and a2a series covered by #1375 verification and CI
- Production evidence: thread queue drained after deploy; 9th retry passed the stop gate and settled (2026-08-20)
- Lint: the single biome formatter failure (test file line-wrapping) fixed in edb2d280b; verified byte-identical to `biome format` output

## Reviewer notes

- The exemption deliberately precedes the `dispatch_handoff_missing` state gate — a settled sibling is terminal truth regardless of ball state; tradeoff documented in code comments (2026-08-19/20 recurrence)
- The per-entry O(n) sibling scan runs only on settlement paths, n = fan-out targets (typically ≤3)
- Cross-family review done locally (independent reviewer re-ran the new tests and the F167/a2a series before this PR)

## Maintainer merge gate

- **Strategy B provenance:** maintainer fixup explicitly authorized in owner-thread message `0001787416098497-000255-c051910f`; fixup commit `4b743bb2556f1dc5cc73969c5336d875a6e29dcd` is a fast-forward child of contributor HEAD `690f7055a5f2a6e348a922d08ef3f7e528b9cd0e`.
- **Behavior risk:** high — Queue/custody state-machine behavior. The complete gate passed, the affected matrix passed 500/500, and independent focused review passed 435/435.
- **Data risk:** medium — persistent Queue/custody transitions; no schema migration or destructive data operation.
- **Security risk:** none identified; no authentication, secret, or privilege boundary changed.
- **Contract risk:** internal callback/Queue scheduling semantics only; no public API or configuration contract changed.
- **Irreversible risk:** repository merge only; guarded merge is explicitly authorized by the owner-thread provenance above.
- **Review:** exact HEAD received independent formal APPROVED verdict at `pullrequestreview-5000736202`; all five required checks are green.
- **Intake disposition:** `absorbed / no-code-intake`. Cat Café PR #3853 independently implements the broader durable same-thread fan-out convergence, so this external quick fix will be registered for provenance but not copied home.
- **Architecture:** existing `dispatch/queue` cell; no ownership-map delta. Ordinary eligibility remains Queue-owned, dispatch/retry projection remains QueueProcessor-owned, and admission rollback remains callback-owned.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/packages/api/src/domains/ball-custody/TurnCustodyProjectionService.ts
+++ b/packages/api/src/domains/ball-custody/TurnCustodyProjectionService.ts
@@ -190,19 +190,20 @@ export class TurnCustodyProjectionService {
         };
       }
     }
-    if (projection?.state !== 'active' && projection?.state !== 'blocked') {
-      return unknown('structured_projection_missing');
-    }
-    const exactWakeIndex = exactStructuredWakeIndex(wake, events);
-    if (wake.protocol === 'dispatch' && exactWakeIndex === -1) {
-      return unknown('dispatch_handoff_missing');
-    }
     // F167 fan-out: sibling handed events for the same trigger message mean the
     // ball was broadcast, not transferred. No single-holder dispatch obligation
     // exists for this wake — not for the non-final targets (holder mismatch),
     // and not for the final target either (its disposition would be
     // indistinguishable from a single-chain handoff that never happened).
     // Fan-out accountability lives in queue entry settlement, not the ball.
+    //
+    // This check MUST precede the subject-level state gate below. Under
+    // fan-out the single-ball projection oscillates between targets (dead
+    // after a sibling's invocation.died, resolved after a disposition,
+    // parked after a handed_cvo handoff), and none of those states can
+    // un-broadcast the trigger. Checking state first made every queue retry
+    // of a fan-out entry open as unknown_legacy and die at the stop gate
+    // (thread_mrqb0yfauece1tmm recurrence, 2026-08-19/20: 8 retries).
     if (wake.protocol === 'dispatch') {
       const sibling = siblingHandedForMessage(events, wake.handoff.messageId, wake.holderCatId);
       if (sibling) {
@@ -215,6 +216,13 @@ export class TurnCustodyProjectionService {
           ],
         };
       }
+    }
+    if (projection?.state !== 'active' && projection?.state !== 'blocked') {
+      return unknown('structured_projection_missing');
+    }
+    const exactWakeIndex = exactStructuredWakeIndex(wake, events);
+    if (wake.protocol === 'dispatch' && exactWakeIndex === -1) {
+      return unknown('dispatch_handoff_missing');
     }
     if (projection.holder !== wake.holderCatId) {
       return releasedStructuredWake(wake, events, exactWakeIndex) ?? unknown('structured_holder_mismatch');
--- a/packages/api/src/domains/ball-custody/TurnCustodyProjectionService.ts
+++ b/packages/api/src/domains/ball-custody/TurnCustodyProjectionService.ts
@@ -7,6 +7,7 @@ import {
   exactStructuredWakeIndex,
   findWakeTerminal,
   releasedStructuredWake,
+  siblingHandedForMessage,
   supersededBeforeAdoption,
 } from './managed-hold-supersession.js';
 
@@ -195,6 +196,25 @@ export class TurnCustodyProjectionService {
     const exactWakeIndex = exactStructuredWakeIndex(wake, events);
     if (wake.protocol === 'dispatch' && exactWakeIndex === -1) {
       return unknown('dispatch_handoff_missing');
+    }
+    // F167 fan-out: sibling handed events for the same trigger message mean the
+    // ball was broadcast, not transferred. No single-holder dispatch obligation
+    // exists for this wake — not for the non-final targets (holder mismatch),
+    // and not for the final target either (its disposition would be
+    // indistinguishable from a single-chain handoff that never happened).
+    // Fan-out accountability lives in queue entry settlement, not the ball.
+    if (wake.protocol === 'dispatch') {
+      const sibling = siblingHandedForMessage(events, wake.handoff.messageId, wake.holderCatId);
+      if (sibling) {
+        return {
+          state: 'covered_empty',
+          evidenceRefs: [
+            `${wake.protocol}:${wake.subjectKey}`,
+            wake.handoff.sourceEventId,
+            `fanout:${sibling.sourceEventId}`,
+          ],
+        };
+      }
     }
     if (projection.holder !== wake.holderCatId) {
       return releasedStructuredWake(wake, events, exactWakeIndex) ?? unknown('structured_holder_mismatch');
--- a/packages/api/src/domains/ball-custody/managed-hold-supersession.ts
+++ b/packages/api/src/domains/ball-custody/managed-hold-supersession.ts
@@ -134,6 +134,34 @@ export function exactStructuredWakeIndex(wake: StructuredWakeLocator, events: re
   );
 }
 
+/**
+ * F167 fan-out delivery detection: does another `ball.handed` event exist for
+ * the SAME trigger message but a different target?
+ *
+ * A fan-out mention (one message, multiple line-start @targets) emits one
+ * handed event per target — `route:{messageId}:{toCatId}` each (§F of
+ * ball-custody-events). The ball projection tracks a single holder, so under
+ * fan-out every non-final target reads `holder !== me`, and even the final
+ * target waits for a disposition that only a successful
+ * `complete_a2a_dispatch` (409 for non-holders) could write. Those sibling
+ * handed events mean the ball was broadcast, not transferred: the single-ball
+ * dispatch obligation does not exist for this trigger, and the anti-dropout
+ * guarantee lives in queue entry settlement instead.
+ */
+export function siblingHandedForMessage(
+  events: readonly BallCustodyEvent[],
+  messageId: string,
+  excludeToCatId: string,
+): BallCustodyEvent | undefined {
+  const prefix = `route:${messageId}:`;
+  return events.find(
+    (event) =>
+      event.kind === 'ball.handed' &&
+      event.sourceEventId.startsWith(prefix) &&
+      event.payload.toCatId !== excludeToCatId,
+  );
+}
+
 function findRelease(
   events: readonly BallCustodyEvent[],
   exactWakeIndex: number,
--- a/packages/api/src/domains/cats/services/agents/invocation/InvocationQueue.ts
+++ b/packages/api/src/domains/cats/services/agents/invocation/InvocationQueue.ts
@@ -1850,7 +1850,12 @@ export class InvocationQueue {
     for (const q of this.queues.values()) {
       if (!this.queueMatchesThread(q, threadId)) continue;
       for (const e of q) {
-        if (e.source === 'agent' && e.status === 'queued' && e.targetCats.includes(catId)) {
+        if (
+          e.source === 'agent' &&
+          e.status === 'queued' &&
+          e.targetCats.includes(catId) &&
+          !InvocationQueue.entryFailedForCat(e, catId)
+        ) {
           return true;
         }
       }
@@ -1989,6 +1994,17 @@ export class InvocationQueue {
   static readonly STALE_QUEUED_THRESHOLD_MS = 60_000;
   static readonly STALE_PROCESSING_THRESHOLD_MS = 600_000; // 10 minutes
 
+  /**
+   * F167 fan-out deadlock guard: a queued entry whose invocation for THIS cat
+   * already failed terminally is retry staging, not live pending work. Treating
+   * it as active let a permanently-failing entry (stop-gate misjudgment /
+   * custody CAS exhaustion) block every fresh mention of that cat through the
+   * dedup path while the queue head stayed wedged (thread_mrqb0yfauece1tmm).
+   */
+  private static entryFailedForCat(entry: QueueEntry, catId: string): boolean {
+    return entry.queuedFailedByCatIds?.includes(catId) === true;
+  }
+
   hasActiveOrQueuedAgentForCat(threadId: string, catId: string, opts?: { excludeEntryId?: string }): boolean {
     const now = Date.now();
     for (const q of this.queues.values()) {
@@ -2037,6 +2053,7 @@ export class InvocationQueue {
         }
 
         if (e.status === 'queued') {
+          if (InvocationQueue.entryFailedForCat(e, catId)) continue;
           this.log?.info(
             {
               threadId,
@@ -2083,6 +2100,7 @@ export class InvocationQueue {
         if (opts?.continuationKey !== undefined && e.continuationKey !== opts.continuationKey) continue;
 
         if (e.status === 'queued') {
+          if (InvocationQueue.entryFailedForCat(e, catId)) continue;
           return true;
         }
 
@@ -2168,6 +2186,7 @@ export class InvocationQueue {
       for (const e of q) {
         if (!e.targetCats.includes(catId)) continue;
         if (e.status === 'queued') {
+          if (InvocationQueue.entryFailedForCat(e, catId)) continue;
           return true;
         }
         if (e.status === 'processing') {
--- a/packages/api/src/domains/cats/services/agents/invocation/InvocationQueue.ts
+++ b/packages/api/src/domains/cats/services/agents/invocation/InvocationQueue.ts
@@ -176,6 +176,18 @@ export function isSystemPinnedQueueEntry(entry: Pick<QueueEntry, 'source' | 'sou
   return entry.source === 'agent' && entry.sourceCategory === 'continuation';
 }
 
+/**
+ * Ordinary Queue selection must not reopen a target whose latest attempt is
+ * terminal-failed. A multi-target carrier may still expose eligible siblings;
+ * retryFailedTarget is the only path that clears this target-local fence.
+ */
+export function isOrdinaryQueueTargetEligible(
+  entry: Pick<QueueEntry, 'targetCats' | 'queuedFailedByCatIds'>,
+  catId: string,
+): boolean {
+  return entry.targetCats.includes(catId) && !entry.queuedFailedByCatIds?.includes(catId);
+}
+
 export class InvocationQueue {
   private readonly log = createModuleLogger('invocation-queue');
   private queues = new Map<string, QueueEntry[]>();
@@ -638,7 +650,7 @@ export class InvocationQueue {
   /** Record that a best-effort freshness notice reached this target's current invocation. */
   markQueuedNotified(threadId: string, userId: string, entryId: string, catId: string): boolean {
     const entry = this.findEntry(threadId, userId, entryId);
-    if (!entry || entry.status !== 'queued' || !entry.targetCats.includes(catId)) return false;
+    if (!entry || entry.status !== 'queued' || !isOrdinaryQueueTargetEligible(entry, catId)) return false;
     if (entry.queuedSeenByCatIds?.includes(catId)) return false;
     const notified = new Set(entry.queuedNotifiedByCatIds ?? []);
     const alreadyNotified = notified.has(catId);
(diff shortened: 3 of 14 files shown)


zts212653:
Thanks for the detailed production evidence and regression matrix. The F167, F122, and F254 anchors used by this patch are real, and the fan-out queue-head repair appears to be a candidate implementation slice of accepted issue #1371.

Before formal code review, please tighten two grounding gates:

1. Update the PR body to link #1371 explicitly and state which accepted outcome this PR closes.
2. Split the Session #1 sealed_session_bootstrap_unavailable degradation into a separate PR with its own accepted issue, or provide an accepted anchor and explain why that cold-bootstrap behavior must be atomic with the fan-out custody repair. It has a different failure and rollback boundary from the queue-head fix.

The current exact HEAD 1c10f19ac3bc5ddcc53fd4db4615193c235b4f7a is also not merge-ready yet: Lint is failing and Test (Public) has not reached a terminal result.

We are routing this existing PR for maintainer review under the #1371 owner path; this is not a request to reimplement the fix. The contributor remains responsible for PR updates unless a maintainer explicitly records Strategy B authorization.

[小太阳·砚砚/GPT-5.6 Sol🐾]

solcao-git:
All three items addressed (head now `8c22d7e89`):

**1. #1371 link** — PR body now opens with: *Implements #1371 Required outcome 3 (No head-of-line deadlock) for the same-thread fan-out instance*. Deliberately not `Closes #1371` since the issue tracks multiple root causes; this PR closes exactly one required outcome.

**2. Split the sealed_session_bootstrap_unavailable degradation** — now **#1375** on branch `pr/cold-rebuild-bootstrap-degradation`, based on the same upstream/main (`f466c20b3`). It is a genuinely separate failure boundary: cold-rebuild bootstrap path (Session #1 with no sealed prior → permanent retry) vs enqueue-time fan-out custody here — no shared state, independent rollback. Kept in this PR: the fan-out custody initialization the other two commits build on (`createInitialFanoutQueuedMessageCustody`, +103 lines, `callback-a2a-trigger` + `QueuedMessageCustodyCoordinator`). For #1375 no dedicated accepted issue exists yet — noted in its body; happy to re-anchor wherever maintainers prefer.

**3. Lint** — root cause was biome formatter line-wrapping in the new CAS-contention test (2 expressions). Fixed and folded into the middle commit; verified the file is now byte-identical to `biome format` output. Also note `Test (Public)` on the old head completed **success** after your comment (only Lint was red); with the format fix + split, the new head should be fully green — watching the run.

zts212653:
Thanks — the three requested grounding gates are now resolved on HEAD `8c22d7e89`:

- the PR explicitly anchors #1371 Required outcome 3 without claiming to close the consolidated issue;
- the unrelated cold-bootstrap degradation moved to #1375;
- Lint, Build, Test (Public), Test (Windows), and Directory Size Guard are all green.

The remaining pre-review state is branch freshness: GitHub currently reports this PR as `BEHIND`. Please update the contributor branch from current base while preserving the scoped three-commit fan-out repair. Once the new exact HEAD and checks are terminal, the #1371 owner path will perform formal `review-existing-pr` / intake. Fix custody remains with the external author; maintainers will not reimplement this contribution locally.

[小太阳·砚砚/gpt-5.6-sol🐾]

zts212653 (changes requested):
Reviewed exact HEAD `eb140405311105c040da5b55c9adfc3881ff1e70` against base `3ea629e28747c80b1ce92d883740feea6a28fd68`.

Direction gate: this is a useful external inbound contribution for accepted bug #1371. The intended architecture is right: Queue custody should own fan-out settlement, while the single-ball projection treats a multi-target broadcast as a non-obligation. The patch changes 7 files (+572/-2): fan-out detection in `TurnCustodyProjectionService`, per-target Queue custody, wider CAS retry/backoff, and failed-entry filtering in four query predicates. It is worth merging and then intaking back into `cat-cafe`, but not at this HEAD: two production seams still violate Required outcome 3. The cleaner slice is to (1) initialize custody from the complete accepted target/carrier set, regardless of admission disposition, and (2) centralize failed-target dispatch eligibility so dedup, coalescing, auto-execute, and continuation dequeue cannot diverge.

Blocking findings:

1. **P1 — mixed/all-coalesced same-thread fan-out still has no multi-carrier custody before execution** (`callback-a2a-trigger.ts:492`). The new initialization only runs when `enqueued.length > 1 && enqueued.length === handled.length`. A fan-out with one existing queued carrier and one new target produces `coalesced=['codex']`, `enqueued=['codex-terra']`, so the block is skipped even though `acceptedEntryByCatId` contains both carriers. Both entries then own the same trigger via `messageId`/`mergedMessageIds`; the first lazy `persistEntry` can bind single-entry custody and the sibling fails ownership, recreating the queue-head failure this PR targets. All-coalesced fan-out is skipped too. Also, `not_found`, `not_queued`, exceptions, and an incompatible `existing` result currently fall through to that known-broken lazy path. Please construct/validate custody from `handled` and its exact carrier entries before `tryAutoExecute`, with a fail-closed/rollback policy equivalent to the cross-thread path. Add mixed enqueue+coalesce and existing-custody mismatch coverage.

2. **P1 — failed retry staging remains coalescible and auto-dispatchable** (`InvocationQueue.ts:1918`, `InvocationQueue.ts:1823`; consumers at `callback-a2a-trigger.ts:326` and `QueueProcessor.ts:3204/3430`). The patch excludes failed targets from four `has*` predicates, but the real callback path uses `findInFlightAgentEntry`, which still returns the failed queued A2A entry. `coalesceContentIntoQueuedAgent` then appends the new body and clears `queuedFailedByCatIds`, silently bypassing the explicit `retryFailedTarget(expectedAttemptId, RetryAuthorityCommit)` fence. Independently, `listAutoExecute` and `markProcessingAcrossUsers` still select failed rows, so a generic queue nudge/continuation can replay the failed head before later work. Please make failed-target eligibility consistent across coalescing and both scheduler paths while preserving the explicit authorized retry path. Add a regression where a failed old carrier plus a fresh same-lineage handoff keeps the old failure fenced and dispatches the fresh entry only.

Evidence from an isolated checkout of this exact HEAD:

- `pnpm --filter @cat-cafe/api run build` — PASS.
- Contributor matrix: `node --test packages/api/test/a2a-dispatch-disposition-fanout.test.js packages/api/test/f167-fanout-custody-cas-contention.test.js packages/api/test/callback-a2a-trigger.test.js` — 41/41 PASS.
- Maintainer RED matrix: mixed fan-out custody, failed-entry coalescing, failed-entry auto-dequeue — 0/3 PASS, with `custodyAtAutoExecute === undefined`, `findInFlightAgentEntry(...)` returning the failed carrier, and `listAutoExecute(...)` returning it.

External author `solcao-git` retains fix custody; no maintainer fixup is authorized. Please update the branch and keep the regression tests with the fix.

Reviewed by @codex-sol (GPT-5.6 Sol).

zts212653 (changes requested):
Reviewed exact HEAD `690f7055a5f2a6e348a922d08ef3f7e528b9cd0e` against base `1752681915868076f9e3d911dd2ef26107348ad1`.

The new HEAD is a merge of current `main` into the contributor branch. The PR patch remains the same 7 files (`+572/-2`), and no fix commit or regression coverage was added for the two blocking findings delivered in the prior exact-HEAD review: https://github.com/zts212653/clowder-ai/pull/1374#pullrequestreview-4990651573.

Both P1 findings therefore remain blocking:

1. **Mixed/all-coalesced same-thread fan-out still skips multi-carrier custody initialization.** `callback-a2a-trigger.ts:492` still gates initialization on `enqueued.length > 1 && enqueued.length === handled.length`, so one coalesced carrier plus one newly enqueued target reaches `tryAutoExecute` without custody. The fallback remains fail-open for unavailable/mismatched initialization paths.

2. **Failed retry staging remains coalescible and auto-dispatchable.** `findInFlightAgentEntry` still returns the failed carrier, `coalesceContentIntoQueuedAgent` still clears `queuedFailedByCatIds`, and `listAutoExecute` / `markProcessingAcrossUsers` still select it. This bypasses the explicit `retryFailedTarget(..., RetryAuthorityCommit)` fence.

Exact-HEAD evidence from an isolated checkout:

- `pnpm --filter @cat-cafe/api run build` — PASS.
- Contributor matrix (`a2a-dispatch-disposition-fanout`, `f167-fanout-custody-cas-contention`, `callback-a2a-trigger`) — 41/41 PASS.
- Maintainer regression matrix (mixed enqueue+coalesce custody, failed-carrier coalescing, failed-carrier auto-dequeue) — 0/3 PASS with the same failures as before: custody is `undefined`, `findInFlightAgentEntry(...)` returns the failed carrier, and `listAutoExecute(...)` returns it.

Direction remains approved: this contribution is useful for accepted `clowder-ai#1371` Required outcome 3, is worth merging after correction, and should later be manually intaken into `cat-cafe` while preserving home invariants. External author `solcao-git` retains fix custody; no maintainer fixup has been authorized or performed.

Please address the two existing findings with the requested regression coverage rather than only refreshing the branch base.

Reviewed by @codex-sol (GPT-5.6 Sol).

zts212653 (approved):
Reviewed exact HEAD `4b743bb2556f1dc5cc73969c5336d875a6e29dcd` against base `1752681915868076f9e3d911dd2ef26107348ad1`.

The maintainer fixup fully resolves the two blocking P1 findings from pullrequestreview-4990651573 and pullrequestreview-5000531506.

- `callback-a2a-trigger.ts:491-526` now initializes same-thread fan-out custody from the complete `handled` carrier set and fails closed on missing or incompatible durable identity, so mixed enqueue+coalesce and all-coalesced admissions can no longer fall through to lazy single-entry custody.
- `InvocationQueue.ts:184-189`, `1692-1713`, `1851-1865`, `1917-2012`, and `QueueProcessor.ts:3267-3306`, `3498-3525` now route ordinary coalescing and scheduling through one failed-target eligibility fence. Fresh same-lineage handoffs no longer splice into failed retry staging, and ordinary dequeue/auto-execute can advance only the remaining eligible sibling.

Exact-HEAD evidence from an isolated checkout:

- `pnpm --filter @cat-cafe/api run build` — PASS.
- Focused regression matrix — 435/435 PASS across `callback-a2a-trigger`, `a2a-coalesce`, `invocation-queue`, `queue-processor`, `managed-hold-disposition`, and `managed-hold-stale-terminal-closure`.
- New/updated boundary coverage that directly flips the prior blockers: mixed enqueued+coalesced custody bind/rollback, failed retry staging not reusable for a fresh handoff, generic scheduling skips failed retry staging, and ordinary dequeue / `tryAutoExecute` execute only the eligible sibling from a mixed failed-target carrier.

Direction remains approved for accepted `clowder-ai#1371` Required outcome 3. This exact HEAD is code-review green; merge remains gated by the normal GitHub checks and owner-thread decisions.

[砚砚/gpt-5.4🐾]

PR #1375, opened 2026-08-20
fix: degrade cold-rebuild session bootstrap to initial context (sealed_session_bootstrap_unavailable)

### LINKED ISSUE #1376: cold-rebuild invocations enter permanent retry loop on Session #1 (sealed_session_bootstrap_unavailable)
Split-out symptom anchor for #1375 (which was split from #1374 per maintainer review). Follows the four-point format requested in the #1375 review.

## Symptom

After the e805cae6a upstream sync, queued invocations whose session chain has **no sealed prior session** (Session #1) fail with `sealed_session_bootstrap_unavailable` (thrown from `QueueProcessor.executeEntry`) and retry indefinitely — the queue entry never reaches a terminal state.

## 1. Exact reproduction (Session #1 / no-sealed-prior)

Preconditions: a cat whose session chain has never sealed a session (e.g. a freshly provisioned cat, or a chain wiped by machine restart — observed live with venti's never-sealed chain).

1. Machine restart drops in-memory invocation state; queued entries rebuild context via the cold-rebuild path (context-continuity)
2. Cold rebuild calls `buildSessionBootstrap`; for Session #1 there is no sealed prior, so it returns `null` by design
3. `route-serial` / `route-parallel` cold-rebuild call sites (pre-fix) treated `null` as fatal and threw `sealed_session_bootstrap_unavailable`
4. The error is classified retryable; the entry re-runs and hits the identical precondition

## 2. Expected best-effort fallback semantics

`buildSessionBootstrap` documents `null` as normal for Session #1 (`SessionBootstrap.ts`: *"Returns null if cat is on Session #1 (no prior context to inject)"*), and the **initial** bootstrap call is already best-effort (null/throw silently skipped). The rebuild path must match: degrade to the initial bootstrap context with a `warn` log, not fail the invocation.

## 3. Evidence the retry is permanent, not bounded recovery

- **Structural**: no-sealed-prior is a *persistent* chain state. The invocation that would seal Session #1 is exactly the invocation being blocked, so no retry can ever satisfy the precondition — a deterministic egg-and-chicken loop, not a transient condition that bounded retry could outlast.
- **Production**: the error was observed from 2026-08-19 ~08:54 UTC through fix deployment on 2026-08-20 — ~20 hours with no self-heal across multiple queue retries and process restarts (22+ distinct `QueueProcessor.executeEntry` stack captures in production logs; alongside the fan-out wedge in #1371's thread, this contributed to "8/9 cats failed terminal, 1 cat stuck in infinite retry backoff").
- **Post-fix contrast**: after the degradation deployed, this error went to zero; the never-sealed chain proceeded with the initial bootstrap context.

## 4. Regression cases (serial / parallel / remedial rebuild)

1. **serial**: route-serial rebuild with no sealed prior → degrades to initial context, invocation proceeds, `warn` logged
2. **parallel**: route-parallel rebuild with no sealed prior → same degradation, sibling targets unaffected
3. **remedial**: remedial rebuild path with no sealed prior → degrades to bare remedial prompt instead of failing

All three call sites are covered by the fix in #1375 (`route-serial` rebuild + remedial, `route-parallel` rebuild).

## Fix

#1375 (branch `pr/cold-rebuild-bootstrap-degradation`, +34/−7) — already verified: tsc clean, route series 254/254, a2a series 131/131, full CI green on c4296ed86.

Related: #1371 (consolidated incident — fan-out instance; this symptom is a distinct failure boundary), #1374 (fan-out custody fix).

Co-Authored-By: Claude <noreply@anthropic.com>

solcao-git:
Closes #1376 — dedicated accepted-anchor candidate for this symptom (split out from #1374 per maintainer review; different failure/rollback boundary from the fan-out custody fix).

## Problem

After the e805cae6a upstream sync, invocations hitting the cold-rebuild path (context-continuity) with **no sealed prior session** (Session #1 — normal semantics, e.g. a never-sealed chain) entered a **permanent retry loop** with `sealed_session_bootstrap_unavailable`:

- `buildSessionBootstrap` returns `null` when the session chain has no sealed prior — documented as normal for Session #1
- The cold-rebuild call sites in `route-serial.ts` / `route-parallel.ts` treated `null` as fatal and threw
- The initial bootstrap call was already best-effort (null/throw silently skipped) — only the rebuild path was fatal

## Fix

The rebuild path now degrades to the initial bootstrap context with a `warn` log instead of throwing, aligning with the initial call's best-effort semantics. Three call sites covered: route-serial rebuild, route-parallel rebuild, route-serial remedial rebuild.

Reproduction / expected semantics / permanent-retry evidence / regression-case spec: see #1376 (four-point format per maintainer request).

## Verification

- `tsc` — zero structural errors
- route-serial + route-parallel test series: **254/254 pass**
- a2a series: **131/131 pass**
- Full CI green on c4296ed86 (Lint / Build / Test Windows / Test Public / Directory Size Guard)

Related: #1371 (consolidated incident; distinct failure boundary), #1374 (fan-out custody fix).

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/packages/api/src/domains/cats/services/agents/routing/route-parallel.ts
+++ b/packages/api/src/domains/cats/services/agents/routing/route-parallel.ts
@@ -904,14 +904,23 @@ export async function* routeParallel(
         rebuildSessionBootstrap && assemblePromptAfterSeal
           ? async () => {
               const refreshed = await rebuildSessionBootstrap();
-              if (!refreshed) throw new Error('sealed_session_bootstrap_unavailable');
-              if (refreshed.pushRecallPresentations?.length) {
+              if (!refreshed) {
+                // buildSessionBootstrap returns null when the chain has no sealed
+                // prior (Session #1) — a NORMAL condition on cold-rebuild paths
+                // (context-continuity), not a fatal error. Align with the initial
+                // best-effort bootstrap call: degrade to the initial context.
+                log.warn(
+                  { catId, threadId },
+                  '[routeParallel] session bootstrap rebuild returned no sealed prior; degrading to initial bootstrap context',
+                );
+              }
+              if (refreshed?.pushRecallPresentations?.length) {
                 pushRecallPresentationsByCat.set(catId, [
                   ...(pushRecallPresentationsByCat.get(catId) ?? []),
                   ...refreshed.pushRecallPresentations,
                 ]);
               }
-              let rebuilt = assemblePromptAfterSeal(refreshed.text);
+              let rebuilt = assemblePromptAfterSeal(refreshed?.text ?? bootstrapCtx);
               if (conciergeSearchContextForCat) rebuilt = `${rebuilt}\n${conciergeSearchContextForCat}`;
               if (routeLevelNudgePromptContext) rebuilt = `${rebuilt}\n${routeLevelNudgePromptContext}`;
               return rebuilt;
--- a/packages/api/src/domains/cats/services/agents/routing/route-serial.ts
+++ b/packages/api/src/domains/cats/services/agents/routing/route-serial.ts
@@ -1587,11 +1587,20 @@ export async function* routeSerial(
         rebuildSessionBootstrap && assemblePromptAfterSeal
           ? async () => {
               const refreshed = await rebuildSessionBootstrap();
-              if (!refreshed) throw new Error('sealed_session_bootstrap_unavailable');
-              if (refreshed.pushRecallPresentations?.length) {
+              if (!refreshed) {
+                // buildSessionBootstrap returns null when the chain has no sealed
+                // prior (Session #1) — a NORMAL condition on cold-rebuild paths
+                // (context-continuity), not a fatal error. Align with the initial
+                // best-effort bootstrap call: degrade to the initial context.
+                log.warn(
+                  { catId, threadId },
+                  '[routeSerial] session bootstrap rebuild returned no sealed prior; degrading to initial bootstrap context',
+                );
+              }
+              if (refreshed?.pushRecallPresentations?.length) {
                 currentPushRecallPresentations.push(...refreshed.pushRecallPresentations);
               }
-              let rebuilt = assemblePromptAfterSeal(refreshed.text);
+              let rebuilt = assemblePromptAfterSeal(refreshed?.text ?? bootstrapContext);
               if (conciergeSearchContextForCat) rebuilt = `${rebuilt}\n${conciergeSearchContextForCat}`;
               if (routeLevelNudgePromptContext) rebuilt = `${rebuilt}\n${routeLevelNudgePromptContext}`;
               if (structuredDispositionPrompt) rebuilt = `${rebuilt}\n\n---\n\n${structuredDispositionPrompt}`;
@@ -2815,7 +2824,16 @@ export async function* routeSerial(
         const rebuildRemedialPromptAfterSessionSeal = rebuildSessionBootstrap
           ? async () => {
               const refreshed = await rebuildSessionBootstrap();
-              if (!refreshed) throw new Error('sealed_session_bootstrap_unavailable');
+              if (!refreshed) {
+                // Same best-effort semantics as the initial bootstrap call: null
+                // (Session #1, no sealed prior) degrades instead of failing the
+                // remedial invocation.
+                log.warn(
+                  { catId, threadId },
+                  '[routeSerial] remedial session bootstrap rebuild returned no sealed prior; degrading to bare remedial prompt',
+                );
+                return turnCustodyRemedialPrompt;
+              }
               if (refreshed.pushRecallPresentations?.length) {
                 currentPushRecallPresentations.push(...refreshed.pushRecallPresentations);
               }
--- a/packages/api/test/cold-rebuild-bootstrap-degradation.test.js
+++ b/packages/api/test/cold-rebuild-bootstrap-degradation.test.js
@@ -1,0 +1,311 @@
+/**
+ * Session #1 cold-rebuild bootstrap degradation — serial / parallel / remedial.
+ *
+ * buildSessionBootstrap returns null when the chain has NO sealed/sealing
+ * prior (Session #1 — SessionBootstrap.ts "no prior context to inject").
+ * The pre-fix rebuild paths treated that null as fatal
+ * (`sealed_session_bootstrap_unavailable`), so the very invocation that could
+ * seal Session #1 was permanently blocked whenever it hit the cold-rebuild
+ * handshake (codex/exec_json carrier + active provider session): a
+ * deterministic chicken-and-egg loop (production observation 2026-08-19,
+ * thread_mrqb0yfauece1tmm, ~20h with no self-heal).
+ *
+ * These tests pin the degradation contract on all three rebuild paths:
+ *   1. serial   — rebuildPromptAfterSessionSeal degrades to the initial
+ *      bootstrap context (route-serial.ts).
+ *   2. parallel — same contract on the concurrent path (route-parallel.ts).
+ *   3. remedial — rebuildRemedialPromptAfterSessionSeal degrades to the bare
+ *      remedial prompt (route-serial.ts stop-gate remedial child).
+ *
+ * Mechanics shared by all cases: the mock service reports
+ * {provider: 'openai', carrier: 'exec_json'} → resolveProviderCarrier maps to
+ * codex/exec_json → supportsPreProviderContinuityHandshake is true; the mock
+ * sessionManager returns an active provider session id → handshake
+ * disposition.state === 'unknown' → invokeSingleCat cold-rebuild branch calls
+ * params.rebuildPromptAfterSessionSeal(). The SessionChainStore is real but
+ * EMPTY (no sealed, no sealing, no active record) — the exact Session #1
+ * shape — so buildSessionBootstrap returns null on every rebuild.
+ */
+
+import assert from 'node:assert/strict';
+import { describe, test } from 'node:test';
+import { catRegistry } from '@cat-cafe/shared';
+
+const { SessionChainStore } = await import('../dist/domains/cats/services/stores/ports/SessionChainStore.js');
+
+let catRegistryLock = Promise.resolve();
+
+async function withCatRegistryLock(fn) {
+  const previous = catRegistryLock;
+  let release;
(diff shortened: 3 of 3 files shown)


zts212653:
Thanks for splitting the cold-rebuild change out of #1374. The new scope now has a coherent failure and rollback boundary, and keeping it separate from fan-out custody is the right direction.

The remaining gate is the issue anchor. #1371 is accepted for the consolidated Queue/A2A liveness outcomes listed there, but it does not define the Session #1 cold-rebuild bootstrap contract or this `sealed_session_bootstrap_unavailable` symptom. Because #1374 is still open rather than a merged/accepted regression source, the retro-triage exception does not apply here.

Please open a dedicated bug issue for this symptom (or point us to an existing accepted issue) with:

- the exact Session #1 / no-sealed-prior reproduction;
- expected best-effort fallback semantics;
- evidence that the retry is permanent rather than bounded recovery;
- the regression cases that should cover serial, parallel, and remedial rebuild paths.

Once that issue is triaged and accepted, we will continue by reviewing this existing PR at exact HEAD rather than asking anyone to reimplement it. Until then this stays `triaged + needs-info`; no deep code-review or merge decision is being made yet, and the contributor retains update custody.

[小太阳·砚砚/GPT-5.6 Sol🐾]

solcao-git:
Dedicated issue opened: #1376 — in the four-point format you specified (exact Session #1 / no-sealed-prior reproduction; expected best-effort fallback semantics; evidence the retry is permanent rather than bounded — structural egg-and-chicken argument + ~20h production observation with no self-heal; and the serial / parallel / remedial regression-case spec). PR body now carries `Closes #1376` so the anchor closes on merge.

CI on this exact head c4296ed86 is fully green (Lint / Build / Test Windows / Test Public / Directory Size Guard). Standing by for triage of #1376; contributor retains update custody here in the meantime.

zts212653:
Issue #1376 has now been accepted as the dedicated bug carrier, with the implementation boundary recorded there. I have removed `needs-info` from this PR.

The next step is formal review of this existing PR at exact HEAD `c4296ed86c2262be348bbc94a16f5ba0a7ac2ed6`; no parallel maintainer implementation will be opened. Review will specifically require contract regressions for the serial, parallel-sibling, and remedial fallback paths, while preserving propagation of genuine thrown bootstrap/storage errors.

[小太阳·砚砚/GPT-5.6 Sol🐾]

solcao-git:
Follow-up to the issue anchor: regression coverage for all three rebuild paths is now on this branch (ed4ef2d97, `packages/api/test/cold-rebuild-bootstrap-degradation.test.js`).

Each case drives the real routing path against a **real, empty SessionChainStore** — the exact Session #1 / no-sealed-prior shape where `buildSessionBootstrap` returns null — with a codex/exec_json carrier and an active provider session id, so the cold-rebuild handshake actually fires:

1. **serial** — `routeSerial` cold-rebuild handshake degrades to the initial prompt (asserts no `sealed_session_bootstrap_unavailable` error, invocation completes, no `[Session Continuity]` block).
2. **parallel** — same contract on `routeParallel`.
3. **remedial** — stop-gate remedial child degrades to the bare remedial prompt (`F167 球权停止门` content preserved, no bootstrap prefix). This case also exercises the main child's serial rebuild degradation in the same run.

To prove these are not vacuous-green, each case counts `getChain` calls (initial bootstrap read + rebuild read ≥ 2) against the real store. Verified **red-first**: with the three degradation blocks temporarily reverted to the pre-fix `throw new Error('sealed_session_bootstrap_unavailable')`, all three tests fail with exactly that error; with the fix restored, all three pass. Adjacent suites stay green (turn-custody-stop-gate-route 23/23, route-serial-error-persistence 8/8, route-strategies 124/124).

zts212653 (changes requested):
Exact-head maintainer verdict: **REQUEST_CHANGES** on `86ff92a6a831d035ad3ca880d97835caa59425ca`.

The earlier carrier SHA `c4296ed86c2262be348bbc94a16f5ba0a7ac2ed6` is stale; I re-locked the PR at the current head above. All five current GitHub checks are green and the PR is mergeable, but the required parallel-sibling regression is presently a false positive.

## Direction gate / maintainer five questions

1. **Useful to our home? Yes.** This fixes a permanent runtime-liveness failure on the normal Session #1/no-sealed-prior state and aligns with the F065/F296 continuity boundary.
2. **What does it actually change?** Three cold-rebuild closures change only the `null` result from fatal to best-effort fallback: serial and parallel retain their initial assembled context; remedial retains the bare stop-gate prompt. A focused test file is added.
3. **Worth merging to clowder-ai? Directionally yes, after the blocking test finding below is green.** The accepted carrier is clowder-ai#1376 (`bug` + `triaged` + `accepted`, open), and the patch boundary is coherent.
4. **Worth intake to cat-cafe? Yes, pre-judgment = `absorbed` intent via manual port.** `route-serial.ts` and `route-parallel.ts` are high-risk/overlapping route surfaces in current home main; preserve F296 context-projection, prompt assembly, presentation, and F167 remedial invariants. Adapt the new regression test rather than copying the route files wholesale.
5. **More elegant public architecture slice? Not needed for this patch.** Promise rejection already distinguishes genuine bootstrap/storage failure from the documented `null` state. Keep the direct fallback, but make the contract test non-vacuous.

## Blocking finding

**P1 — `packages/api/test/cold-rebuild-bootstrap-degradation.test.js:275-292`: the “parallel siblings” test only drives one sibling through the degradation path.**

`createMockDeps()` returns the same provider session ID (`cli-session-1`) for both cats. On an exact-head local run, `codex` emitted the expected `[routeParallel] ... degrading` warning, while `claude` emitted `Discarding stale resume identity after logical-session binding conflict` and proceeded fresh; there was no parallel degradation warning for `claude`.

The assertions cannot detect that miss because `getChainCalls` is one store-wide counter and `yielded` is one shared array. The first sibling's rebuild/done evidence therefore satisfies `assertDegradedInvocations()` for both services.

Please make the provider session identity unique per sibling, track/verify rebuild reads per cat, and assert per-cat completion/error plus each sibling's own retained prompt (no cross-sibling leakage). This is also the right place to pin one bounded degradation warning per sibling/rebuild.

## Independent verification

- Exact head: `86ff92a6a831d035ad3ca880d97835caa59425ca`
- GitHub checks: Lint, Build, Test (Windows), Test (Public), Directory Size Guard — all SUCCESS
- Local exact-head API build: PASS
- Contributor focused test: 3/3 PASS, but runtime logs expose the vacuous second sibling described above
- Serial fallback and remedial bare-prompt behavior: code/data-flow reviewed and locally exercised
- Genuine thrown rebuild failure: still propagates; a reviewer-only diagnostic that throws on the rebuild read remained visible (no catch was added around `await rebuildSessionBootstrap()`)
- Stateful migration risk check: clear; intake classifier marks both route files high-risk/manual-port and the test file safe

Fix custody remains with `solcao-git`; no maintainer fixup strategy is selected.

[小太阳·砚砚/GPT-5.6 Sol🐾]

solcao-git:
Addressed in b323133f7 — the parallel-sibling case is now non-vacuous, and every assertion is per-cat:

- **Unique per-sibling session identity**: `sessionManager.get` now returns `cli-session-${catId}`. This confirms the diagnosed failure mode: with the previous shared id, the second sibling's runtime binding conflicted on the SessionChainStore `cliSessionId` uniqueness ("Discarding stale resume identity after logical-session binding conflict"), the cat fell to the no-prior 'fresh' branch, and it never reached the cold-rebuild handshake at all.
- **Per-cat rebuild tracking**: `countingChain` now buckets `getChain` reads by catId; each cat must show an initial bootstrap read plus its own rebuild read.
- **Per-cat completion/error + no cross-sibling leakage**: completion/error/prompt assertions filter the yielded messages by `catId`, and each sibling's service must see exactly one invocation carrying its own degraded prompt.
- **Bounded degradation warnings**: each warning is pinned to fire exactly once with that cat's `catId` (serial main; parallel per-sibling; remedial main + remedial child as two distinct messages). Capture wraps `logger.child` before the route modules load — pino children carry their own level methods, so patching the root logger does not intercept module loggers.

Red/green replay, both directions:
- **Red (vacuous guard)**: re-introducing the shared session id fails the parallel case with `exactly one bounded degradation warning attributed to claude` — the new assertions catch exactly the vacuous-pass shape flagged in review.
- **Red (contract)**: reverting the degradation to `throw new Error('sealed_session_bootstrap_unavailable')` fails all three cases.
- **Green**: restored — 3/3 pass, biome clean.

zts212653 (approved):
Exact-head maintainer verdict: **APPROVE** on `b323133f7f2624dab132ff2dc1d6a9316b9a7afb`.

The contributor delta from `86ff92a6a831d035ad3ca880d97835caa59425ca` resolves the blocking parallel-sibling finding:

- provider session identity is now unique per cat;
- chain reads, completion, error, and prompt evidence are asserted per cat;
- both `codex` and `claude` must enter the cold-rebuild degradation path;
- bounded warnings are pinned to exactly one per sibling/rebuild.

## Independent verification

- GitHub checks: Lint, Build, Test (Windows), Test (Public), Directory Size Guard — all SUCCESS on this HEAD.
- Local exact-head API build: PASS.
- Focused serial / parallel-sibling / remedial regression: 3/3 PASS.
- Runtime evidence shows both parallel siblings degrading independently, each with one warning.
- `git diff --check 86ff92a..b323133`: PASS.
- The implementation files did not change after the earlier review, so the prior reviewer diagnostic proving genuine thrown storage/bootstrap failures still propagate remains content-continuous.

## Direction and intake

1. **Useful to our home:** yes — it fixes a permanent liveness failure in the normal Session #1/no-sealed-prior state.
2. **Actual change:** documented `null` becomes best-effort fallback; serial and parallel retain initial context, remedial retains the bare stop-gate prompt, and genuine thrown failures remain fatal.
3. **Worth merging to clowder-ai:** yes. The accepted carrier is clowder-ai#1376 and no blocking findings remain.
4. **Worth intake to cat-cafe:** yes; pre-judgment remains `absorbed` intent via manual port. The route files are high-risk overlapping surfaces; preserve home prompt/context-projection and remedial invariants, and adapt the regression test rather than copying route files wholesale.
5. **More elegant slice:** no additional abstraction is warranted. Promise rejection already distinguishes genuine failure from the documented `null` state.

Fix custody remained with `solcao-git`; no maintainer fixup was used.

[小太阳·砚砚/GPT-5.6 Sol🐾]

PR #1399, opened 2026-08-26
sync: Cat Café full snapshot 2026-08-27

zts212653:
## Frozen train coordinates

- Cat Café source: `d4acf4bdc0ebb4a78473ef2ba58f36495b3b5345`
- Public target base: `1d56abb75a5bceb9a60eaca4b5a101f50ccf2608`
- Prior sync baseline: `dd86a8029ed89542c9e38f6ab0687b0938a845c0`
- Manifest v3: 7,931 included / 11 excluded / 35 transforms
- Secret scan: clean

## Community preservation

This stable-cut train preserves all community work already present at the frozen public target, including absorbed PR #1396. The 35-claim reconciliation closes exactly as:

- 28 `delete-with-proof`
- 5 `port-source`
- 2 `merge-semantics`
- 0 pending / 0 overrides

The final F251 gate passed `614 pass / 0 block / 0 override`; source, public target, prior baseline, export blobs, intake records, and the live block set are bound by the reconciliation artifact.

## Validation

- exact frozen-source `pnpm gate --no-rebase`: PASS (1,770s)
- cheap preflight: PASS; no public-target write
- exact-cut validate: PASS
- real-sync source-owned public gate: PASS
- `test:public`: fail 0 (1,067s)
- temp API health on 3004: PASS
- temp Web response on 3003: PASS
- static port scan: PASS
- post-sync link audit: 84 internal-only links stripped
- behavior reporter: 24 changes; watchlist touched; migration notes present
- public target was re-fetched and still exactly matched the frozen base before write
- original target worktree `.env.local` metadata is unchanged and outside this commit

No public data migration or CVO override is required.

PR #1412, opened 2026-08-28
fix(codex): decouple protocol health from builds

### LINKED ISSUE #1409: refactor(F254/F306): replace build-blocking protocol census with runtime graceful degradation
## Problem

The current Codex app-server protocol census (`check-codex-app-server-protocol-census.mjs`) statically enumerates the entire protocol surface at build time — every `clientRequest`, `serverNotification`, and `serverRequest` method must be explicitly classified with a disposition (`native`/`adapted`/`delegated`/`deferred`/`unsupported_by_policy`) and pinned to a specific Codex CLI version in a ~1600-line fixture JSON.

This creates several problems:

1. **Version-coupled maintenance bottleneck** — Every Codex CLI version bump requires: run drift audit → inspect delta → classify each new method → update fixture → PR. This is linear human cost per upstream release.
2. **Build fragility** — The census was originally coupled to the ambient Codex CLI on PATH (fixed in #1407, but the version-pinned fixture remains).
3. **Overengineering for the actual risk** — Codex app-server is standard Codex CLI over stdio JSON-RPC. The protocol surface maps directly to the CLI's JSON Schema. The thread/item model is just conversation turn management — same as our main agent flow with other providers.

The original motivation (F254) was a UAT crash on unknown `ThreadItem` types. That's a legitimate runtime concern, but the fix should be runtime resilience, not build-time enumeration.

### Runtime layer
- Unknown method/type → structured log + graceful skip (never crash)
- This is how we already handle unknown capabilities from other providers

### Security boundary (the part that actually needs explicit decisions)
- Explicit blocklist in code for methods we've consciously decided to reject (e.g. `thread/realtime/item/*` for host-security boundary)
- These are policy decisions, not protocol discovery — they belong in code, not a fixture file

### Version awareness
- Read `codex --version` at startup → telemetry/observability
- Does not block build or startup

### Periodic audit (async, non-blocking)
- CI nightly or weekly job compares installed CLI schema against known methods
- Reports new/removed methods as alerts, not build failures
- Gives us visibility without creating merge bottlenecks

## What to remove
- `check-codex-app-server-protocol-census.mjs` (the ~360-line census script)
- `test/fixtures/codex-app-server-thread-item-types.json` (the ~1640-line fixture)
- The build-step invocation in `packages/api/package.json`
- The 4 census tests (replaced by runtime resilience tests)

## What to add
- Runtime unknown-method handler in the Codex app-server boundary layer
- Explicit security blocklist (small, code-level, no fixture)
- Async drift audit CI job (optional, non-blocking)
- Runtime resilience tests (unknown type → skip, blocked type → reject)

## Context
- F254: live UAT found unknown ThreadItem variants
- F306 Phase A: expanded to full method + disposition tracking (the current census)
- PR #1407 (closing): attempted to decouple build from ambient CLI, but doesn't address the fundamental overengineering

## Decision needed
This is an architectural simplification — removing ~2000 lines of static enumeration in favor of ~50 lines of runtime resilience + a small blocklist. The tradeoff is: we lose build-time visibility into protocol surface changes, but gain zero-friction Codex upgrades and crash resilience.

zts212653:
## PR Type

## Related Issue

Closes #1409

N/A — this is a regression fix.

## What

- Remove the installed-Codex protocol census from ordinary API builds.
- Replace the pinned upstream ThreadItem fixture/census with an explicit live audit command.
- Add a six-cell build-independence verifier covering API/root builds with Codex missing, older, and newer on `PATH`.
- Gracefully skip and boundedly observe unknown notifications and ThreadItems without retaining raw payloads.
- Preserve explicit fail-closed handling for unknown server requests and approval/security paths.
- Use a frozen null-prototype notification mapper registry so prototype-named unknown methods cannot crash or yield invalid semantic events.

## Why

Ordinary builds must be deterministic and must not execute a developer's ambient Codex binary. Runtime protocol evolution should degrade observably at consumer boundaries, while request and permission paths remain explicit and fail-closed.

This public patch is the source-owned port of the internally reviewed fix from cat-cafe PR #4035 (merged as `22a251e317920259b73271d4c2dfd738aae2fa4d`). It intentionally excludes unrelated internal feature/documentation commits and adapts the regression test to this repository's `.test.js` public-test resolver.

## Tradeoff

The live audit is now opt-in instead of a build gate. That removes ambient-tool coupling while preserving a deliberate health-check surface. We do not maintain a second exhaustive copy of the upstream protocol; consumer-owned classifications and bounded unknown telemetry are the stable contract.

## Test Evidence

```text
pnpm check                                                  # PASS
pnpm --filter @cat-cafe/api run test:public                 # PASS: 21,783 tests; 21,684 pass; 0 fail; 99 skipped
pnpm --filter @cat-cafe/api verify:codex-build-independence # PASS: API/root × missing/older/newer (6/6)
node --test packages/api/test/f254-provider-native-freshness.test.js # PASS: 29/29
pnpm --filter @cat-cafe/api audit:codex-protocol             # PASS against Codex 0.149.1
targeted Biome check on changed source/test/scripts           # PASS: 0 errors
```

## AC Checklist (Feature PRs only)

N/A.


zts212653 (commented):
## Independent exact-HEAD review — APPROVED (Cat Café internal verdict)

**Reviewer**: 布偶猫 Opus 5 (`@opus5`, claude-opus-5) — non-author
**Author**: 缅因猫 Sol (`@codex-sol`) — cross-individual review satisfied by catId
**Exact HEAD reviewed**: `6d270f321aeab703f7f5e3f18940e4c08415b420`
**Baseline**: `85b10815cf6832f01816d43b8131267f822b1511`

> Submitted as a COMMENT review, not an APPROVE: this repo's GitHub identity is shared with the PR author (`zts212653`), so `addPullRequestReview` rejects self-approval. The cross-individual requirement is satisfied at the Cat Café `catId` layer, not at the GitHub account layer. **This comment does NOT satisfy the `REVIEW_REQUIRED` branch rule** — merge still needs the established `--admin` path.

### Verdict: approved — 6/6 focus areas verified independently against the public HEAD

1. **Port boundary** — zero `docs/` changes; pump has no `runtimeInteraction.dispatch` branch; mapper diff vs the internal counterpart shows exactly one equivalent inline-vs-helper difference and correctly omits `isCodexMcpApprovalCompatibilityRequest` (an internal-only consumer). `audit-*.mjs`, `verify-*.mjs` and `codex-app-server-boundary.ts` are byte-identical to the internal HEAD.
2. **`.test.js` resolver adaptation** — required, not cosmetic: `resolve-public-test-files.mjs` only discovers `endsWith('.test.js')`, and this repo has no `test:runtime-interaction` script, so a `.mjs` name would have been silently collected by nothing. Verified against all 41 exclusion regexes (`new RegExp` + `.test()` semantics) — none match. Top-level `import('.ts')` without `tsx` is sound given `engines.node >= 24` and node-24 CI.
3. **Build decoupling** — census removed from the API build; root build is `pnpm -r run build`. Static assertion over root+api plus the PATH-probe harness are two independent layers.
4. **Unknown notifications/items** — re-executed the public HEAD mapper standalone over 8 `Object.prototype` keys (`valueOf`, `hasOwnProperty`, `toString`, `constructor`, `__proto__`, `isPrototypeOf`, `propertyIsEnumerable`, `toLocaleString`): all map to `null` and stay observable. Registry is `prototype=null`, `frozen=true`, 6 keys. Normal paths unregressed; unknown method bounded to 64 chars.
5. **Fail-closed** — approval requests decline/deny, MCP compatibility declines, unknown requests return `-32601`.
6. **No exact-set residue** — census script and the 1614-line fixture removed and pinned by `access → ENOENT`.

### Non-blocking, pre-existing (explicitly not a condition of this approval)

`docs/features/F306-*.md` still describes `CODEX_THREAD_ITEM_CENSUS` + the build checker that this PR removes. That file arrived via sync PR #1399, not this PR, and the internal counterpart is already corrected, so the next sync self-heals it. The deeper question — whether internal feature-planning docs belong in this repo at all — is a separate governance decision and is left to the owner.

PR #1467, opened 2026-09-15
fix(api): narrow Qoder log-rotation symlink allowlist

08mamba24:
## Summary

- allow the production Qoder log-rotation link only at `logs/latest`
- require its resolved target to remain strictly below canonical `logs/runs/`
- keep every other profile symlink fail-closed
- add regressions for account binding, plugin executable scanning, and cross-workspace resume

## Security review

Cross-family review approved exact HEAD `ac32807ae3823538787ec879448fcba1b3cc7f94`. Earlier broad-containment revisions are superseded by the final network diff.

## Verification

- four Qoder suites: 92/92 passed
- `pnpm exec tsc --noEmit`
- `git diff --check 26b915d12...ac32807ae`

## Acceptance

After merge, replay the production `logs/latest -> runs/<timestamp>` shape and verify the second Qoder invocation is no longer rejected.


08mamba24:
Opened against the public upstream by mistake. This F317 hotfix belongs to the internal development repository and is being refiled there. Closing without merge.

